PR #363, opened 2026-04-08
feat(rag): persistent, modular RAG-based tool indexing and retrieval (v3.10.0)

1Jamie:
This PR was originated by @1Jamie — thank you for the foundational work on the RAG tool indexing pipeline! I took over the branch to finish, review, and land the implementation.

## Summary

- **RAG tool indexing pipeline** — ported persistent, modular tool indexing and retrieval system; zero-lag startup, multi-provider tool aggregation, dynamic relevance-based retrieval
- **Multi-intent decomposition** — agent correctly decomposes and retrieves tools for multi-intent queries
- **Tool Index Status sensor** — new diagnostic sensor (`sensor.tool_index_status`) reporting indexing state
- **Actuation safety net** — deduplication and guard logic prevent double-execution at retrieval time; `tool_index_failed` guard prevents crashes when index is unavailable
- **`safe_convert` utility** — wraps `voluptuous_openapi.convert` with robust serialization; handles unhashable selectors, extracts `SelectSelector` options into OpenAPI enums
- **Reasoning disable fix** — corrected reasoning model disable path
- **`confirm_sensitive_action` removed from actuation tools** — it's a PIN handshake, not a physical actuation tool
- **Startup race condition** — logged TODO for tool index readiness race at startup (not yet fixed)

## Key fixes

- Tool retrieval limit enforced as hard cap on total tools at execution time (not just at retrieval)
- Fallback tool selection respects limit when index is not ready
- Door lock no longer fails when PIN is enabled but never configured
- Blind `except Exception` catches replaced with typed exception tuples (`InvalidNamespaceError`, `psycopg.OperationalError`, `psycopg.ProgrammingError`, `ValueError`)

## Test coverage

- `test_tool_retrieval.py` — RAG pipeline, deduplication, safety net, config fallbacks
- `test_conversation_units.py` — agent graph unit tests
- `test_unhashable_selector.py` — unhashable selector serialization edge cases

## Documentation

- **CHANGELOG.md**: v3.10.0 entry
- **README.md**: semantic tool retrieval feature, `tool_retrieval_limit`/`tool_relevance_threshold` config section, Diagnostic Sensors subsection, four-node graph description
- **TODOS.md**: marked `sentinel_camera_entry_links` and weekly/DOW baseline patterns as done

Closes #359


1Jamie:
okay, sorry for the delay, have been a bit busy the last few days, but if you can @goruck take a look at the changes, i have addressed the points in which you referenced and came up with the most robust but efficient solution i could with the constraints you laid out. would love your input. i did base the update on your latest code so its mostly drop in place in regards to the main branch.

goruck (commented):
Thanks for taking this on — the architecture is genuinely on the right track and the SHA256 change detection is a nice touch. But there are a few bugs that block merging, plus some missing pieces from the original issue spec. Walking through them:

---

**Three bugs that need fixing before merge:**

**1. Config options read from the wrong dict level (`graph.py`)**

In `_retrieve_tools`, `opts` is set to `config.get("configurable", {})`. But user options are nested at `config["configurable"]["options"]` — that's how `_call_model` reads them. As written, `CONF_TOOL_RETRIEVAL_LIMIT`, `CONF_TOOL_RELEVANCE_THRESHOLD`, and `CONF_LLM_HASS_API` always fall back to hardcoded defaults. The sliders do nothing.

Fix: `opts = config.get("configurable", {}).get("options", {})`.

**2. Local LangChain tools are permanently excluded (regression)**

Tools indexed under `api_id="hga_local"` — `get_and_analyze_camera_image`, `upsert_memory`, `alarm_control`, `confirm_sensitive_action`, `get_entity_history`, `resolve_entity_ids`, `write_yaml_file`, `add_automation` — can never pass the gatekeeper filter `if api_id in active_api_ids`. `"hga_local"` is not a HA LLM API ID and will never appear in `CONF_LLM_HASS_API`. Same issue affects `ACTUATION_LANGCHAIN_TOOLS`: `alarm_control` and `confirm_sensitive_action` are listed in the safety net but blocked by the same filter. On this PR, the agent silently loses all its specialized tools.

Fix: either always append `"hga_local"` to `active_api_ids` inside `_retrieve_tools`, or handle local tools via a separate always-include path outside the gatekeeper.

**3. Duplicate tool entries for non-`assist` providers**

`add_to_binding` only prevents re-adding a tool if `routing_map[tool_name] == LLM_API_ASSIST`. A tool from a custom provider that appears in both `top_results` and the actuation safety net scan gets appended to `selected_tools` twice. OpenAI and Anthropic both reject requests with duplicate tool definitions.

Fix: `if tool_name in routing_map: return` — deduplicate regardless of provider.

---

**Two missing pieces from the issue spec:**

**Cold-start fallback**: Issue #359 explicitly calls for falling back to the full tool list before embeddings are ready. Currently, when `tool_index_ready` is False, the background task fires but `_retrieve_tools` immediately runs against an empty store and returns zero tools. First-request behavior is worse than today.

Fix: if `selected_tools` is empty after retrieval (cold start, threshold too high, or store empty), fall back to binding the full tool set.

**Graceful degradation**: Same issue — no fallback if `store.asearch` raises (pgvector down, connection error, etc.). Should catch exceptions in `_retrieve_tools` and fall back to full tools rather than propagating up to the graph.

---

**A few smaller things worth fixing:**

- `async-timeout==4.0.3` in `manifest.json` — the PR description says this is a test runner fix. It shouldn't be in the production manifest; it'll force-install a deprecated library on every user's HA instance. Move it to test dependencies only, or use `asyncio.timeout` directly (Python 3.11+, and HGA runs on 3.14).

- `strings.json` and `translations/en.json` are missing entries for `tool_retrieval_limit` and `tool_relevance_threshold`. Without them, the HA UI displays raw key names for the sliders.

- `_run_tool_index_background` sets `tool_index_ready = True` in `finally`, so a failed index run is silently marked complete. Move that line into the `try` block so a failed run retries on the next request.

- Local tool hash (`conversation.py` ~line 356) omits the parameter schema from `raw_content`, so schema changes to local tools are never detected. The provider loop correctly includes the schema in the hash — local tools should do the same.

- The actuation safety net fires a second full `store.asearch(limit=100)` on nearly every query. `ACTUATION_KEYWORDS_REGEX` matches `turn`, `set`, `run`, `play`, `stop`, `start`, `open`, `close` — extremely common words in home automation context. Worth pre-tagging actuation tools during indexing and fetching them deterministically instead.

---

The core concept is solid. Fix the three critical bugs + add the cold-start fallback and it's close to mergeable.

1Jamie:
and this ladies and gentleman is why you verify you coppied the right folder into your test environment XD
im on it, my apologies, holy buttcheeks im blind XD

1Jamie:
okay, here we go @goruck lol, i have gone through and properly validated and tested everything. 
quick note in regards to the dep piece. since it wasn't anything too crazy, i just went through and removed the usage in the actual system of the depreciated ```async_timeout``` and properly replaced the places it was being called in ```tools.py```, ```utils.py``` and ```video_analyzer.py``` so they are using ```asyncio.timeout``` instead. everything is up to current standards of timeout usage.

now for the list!
---
## Bugs:

Bug 1: in ```graph.py``` the new ```_get_allowed_api_ids``` and ```_get_rag_retrieved_tools``` functions properly navigate down into the configuration. we ensure the retrieval limits and thresholds are actually found and followed instead of falling back to defaults.

Bug 2: the function ```_get_allowed_api_ids``` specifically unions the apis with the local id: ```return set(active_api_ids) | {"hga_local"}``` this is done to make sure internal tools are evaluated by the gate keeping pieces and are properly put through the semantic check.

Bug 3: i added ```format_and_dedupe_tools``` to handle conflict resolution "properly". it uses ```if name in routing_map: continue`` to make sure there is a strict first come first serve in the de duplication across all providers. this does not just use the first one it finds though, it is picking the one that best semantically matches and surfacing that one as primary and should solve the issues with the openai and anthropic rejections while also surfacing the best matching tool.

---
## Missing features:

Cold start fallback:
changed ```_get_rag_retrieved_tools``` and ```_get_actuation_safety_tools``` to immediately check ```config.get("configurable", {}).get("tool_index_ready", True)``` so that if the index isnt ready (or it didnt find any) ```_retrieve_tools``` falls back to a new ```_get_fallback_tools``` which assembles and builds the big choker as its fallback providing the complete set instead.

## Graceful Degradation:
The vector store calls are now wrapped in try|except blocks. if the db connection drops or something fails outside of hga's control it does not fully hard fault and allows the graph to keep flowing.

---
## Tweaks and Optimization:

```async-timeout``` Depreciated: as stated in the first paragraph i went ahead and removed the dep and updated the other pieces that were using the older libs to instead use ```asyncio```.

Missing UI Strings: both ```strings.json``` and ```translations/en.json``` now have the keys to properly populate the ui instead of using the keys.

```tool_index_ready``` moved so a failure does not trigger false success.

Local Tool Hash Schema: it now extracts ```t_func.args_schema.schema()``` and appends it to the raw_content before hashing it. should properly handle local tool parameters being changed and properly reindex.

Actuation Safety Ned Optimization: the expensive limit=100 was replaced, now instead it evaluates ```is_actuation_tool(name)``` at indexing and stores it in the metadata. then it uses ```store.alist(..., filter={"is_actuation": True})``` to fetch it deterministically.

Graph Topology Restructure: I broke down the massive nightmare that was ```_retrieve_tools``` and broke it down into isolated helper functions (```_get_rag_retrieved_tools```, ```_get_actuation_safety_tools```, ```_get_fallback_tools```, and ```_format_and_dedupe_tools```) so that not only is it cleaner and easier to maintain... the lint system hushes up lol. though the main point was to add a bit more modularity so that i could wire up the cold start and database error fallbacks without turning the whole node into friday night spaghetti.



i think i checked every box we were looking at, though let me know if you want me to add any more test cases to make sure none of this falls through the cracks later.

(yeah, im bad at markdown stlying....)

**Review by goruck — COMMENTED** [review]
Good work addressing all 10 items from the first review — the refactor into `_get_rag_retrieved_tools`, `_get_actuation_safety_tools`, `_get_fallback_tools`, and `_format_and_dedupe_tools` is clean and easy to follow. Two blockers before merge, one regression worth fixing, and a test gap on the most important bug fix.

---

**Blocker: `make lint` fails**

Three files need reformatting and there are two unused imports + one bare-`Exception` catch in `graph.py`:

```
# graph.py:41-42 — imported but not used (moved to helpers.py)
ACTUATION_LANGCHAIN_TOOLS
ACTUATION_TOOL_PREFIXES

# graph.py:630 — BLE001
except Exception:   # in _get_actuation_safety_tools inner fallback

# helpers.py — unsorted import block + missing blank line between format_tool and is_actuation_tool
```

Run `make fix && make format` and it should clear cleanly.

---

**Blocker: `make typecheck` must pass**

Please run `make typecheck` and fix any pyright errors before merge. The new `RawTool` TypedDict, `MultiLLMAPI`, and the `_retrieve_tools` refactor all introduce new typed surfaces — pyright catches things ruff won't.

---

**Regression: silent API load failure**

The old code returned a user-facing error if `async_get_api` raised. The new loop logs a warning and continues. If a user's configured API fails to load entirely, the agent silently operates with zero HA tools — no message in the UI, no indication anything is wrong. Worth adding a check: if `active_apis` is empty after the loop, return an error response the same way the old code did.

---

**Test gap on the most important bug fix**

Bug #2 (local tools excluded) was the subtlest of the three critical bugs. The fix (`set(active_api_ids) | {"hga_local"}`) is a one-liner that's easy to regress. There's no test that verifies `hga_local` tools survive when the user's `llm_hass_api` config doesn't include `"hga_local"`. A single test covering this path would lock it in permanently.

The three existing tests all exercise the fallback path — good coverage there. The happy path (RAG returns results that pass the score/api_id filter, actuation safety net fires on keyword match) and the deduplication logic have no coverage. Those plus `MultiLLMAPI` routing and `strip_for_embedding` / `truncate_for_embedding_index` can go in a follow-up PR if you'd rather not hold this up.

Fix the lint, typecheck, the silent failure regression, and add the `hga_local` test and this is ready to merge.

**Comment by 1Jamie** [comment]
nah, you know what they say lol, if you are going to do it you might as well do it right. 

issues 1: ```make lint && make typecheck``` resolved, ```ACTUATION_LANGCHAIN_TOOLS``` and ```ACTUATION_TOOL_PREFIXES``` imports were removed from ```graph.py`` as what needed them is in ```helpers.py```. 
though i will clarify i also dealt with the blind exceptions in the graph by making sure it goes through exception tuples instead of just a blind catches that were in use in there too.

issue 2: went through each thing from the typecheck and resolved them, they were relatively minor but good catch, should have ran that and didnt. will make sure to put it on my list of things to check every time.

issue 3: the agent now returns a descriptive error if no LLM APIs can be loaded, clearly listing the failed providers (e.g., ```assist```, ```hga_local```).

issue 4: ```test_tool_retrieval.py``` should cover all cases i could think of and you mentioned:
Fallback on empty store.
Fallback on general Exception.
Fallback when index is not ready.
hga_local survival in allowed IDs.
RAG happy path (successful retrieval).
Actuation Safety Net trigger.
First-seen wins deduplication.
Store is None handling.
Specific exception handling (psycopg, ValueError).


as far as i can tell and from what you have said this should just about wrap up the things you have pointed out unless anything i adjusted introduced new issues. though as far as i can tell and have reviewed everything lints, tests and typechecks properly, doesnt crash when i run it and does what it should from my testing. My preference was to just iron out all the pieces so when it gets merged its the complete system and doesnt need to be come back to just to fix left behind things, instead just for improvements and optimizations. back to you @goruck let me know! :D

**Review by goruck — COMMENTED** [review]
Great effort on this — the architecture is solid and the persistence in working through the review rounds is appreciated.

---

Integration test #1 (cold start) crashes immediately. Two other issues also need fixing before merge.

---

**Bug: Cold start crashes with `PydanticInvalidForJsonSchema`**

On first boot, `_get_fallback_tools` in `graph.py` calls `lc_tool.args_schema.schema()` on tools that have `store: Annotated[BaseStore, InjectedStore()]` in their signature. Pydantic v2 can't generate JSON schema for `IsInstanceSchema(BaseStore)` and raises `PydanticInvalidForJsonSchema`, crashing the entire request.

The same gap exists in the `conversation.py` indexing path — it has a try/except but only catches `(AttributeError, TypeError, ValueError)`, missing this exception.

Fix in `agent/graph.py` — wrap the bare call in `_get_fallback_tools`:
```python
params = "{}"
if hasattr(lc_tool, "args_schema") and lc_tool.args_schema:
    try:
        params = json.dumps(lc_tool.args_schema.schema())
    except (
        AttributeError,
        TypeError,
        ValueError,
        PydanticInvalidForJsonSchema,
    ):
        params = "{}"
```
Add `PydanticInvalidForJsonSchema` to the pydantic import in both files, and add it to the existing except tuple in `conversation.py` (~line 580). Falling back to `"{}"` is correct here — `store` is an injected dependency the LLM shouldn't see anyway.

---

**`make typecheck` still fails — 2 errors**

`test_tool_retrieval.py:306` and `:312` pass `None` to `_get_rag_retrieved_tools` and `_get_actuation_safety_tools`, but both are typed `store: BaseStore`. Both functions already have `if store is None:` guards internally — the signatures just need to match:

```python
async def _get_rag_retrieved_tools(
    store: BaseStore | None, ...

async def _get_actuation_safety_tools(
    store: BaseStore | None, ...
```

---

**Three debug files accidentally committed**

`typecheck_output.txt`, `typecheck_output_v2.txt`, `typecheck_output_v3.txt` are in the repo root (~350KB each). Remove with `git rm typecheck_output*.txt` and consider adding `typecheck_output*.txt` to `.gitignore`.

---

Once those are fixed, here's the integration test checklist to run on HA:

**P0 — must pass**

1. **Cold start / first boot** — clear `tool_content_hashes` from `HGAData` and restart HA. On the first request ask something simple (*"What lights do you have?"*). Should respond correctly using the full tool fallback while the index builds in the background. Watch logs for `tool_index_ready = True` being set after the background task completes. Second request should use RAG retrieval.

2. **Basic HA tool retrieval** — *"Turn on the living room lights."* Expected: `light.turn_on` bound and called. Failure mode: agent says it can't control devices.

3. **Local tools always available** — *"Show me the last camera event in the front yard"* (triggers `get_camera_last_events`) and *"Remember that I prefer Celsius"* (triggers `upsert_memory`). These use `api_id="hga_local"` which is never in the user's configured HA API list — this is the regression test for bug #2.

4. **Actuation safety net** — *"Lock the front door"* and *"Arm the alarm."* Expected: `lock.*` / `alarm_control` tools bound even if they score below the relevance threshold.

**P1 — config slider validation**

5. **Retrieval limit** — set `Tool retrieval limit` to 1, ask *"Turn on the lights and check the temperature."* Watch logs to confirm only 1 tool is retrieved per turn.

6. **Relevance threshold** — set to `0.99` (near-impossible to pass), ask *"Turn on the kitchen light."* Expected: threshold filters everything → fallback to full tool list → still works. Then set to `0.01` and verify normal behavior.

**P2 — multi-provider and error handling**

7. **One provider fails to load** — temporarily misconfigure one API (bad key or URL), restart HA, make a request. Expected: user sees an error message listing which API failed, not silent failure. Logs should show `Could not load LLM API: <api_id>`.

8. **Re-index after restart** — restart HA a second time with no tool changes. Expected: no index tasks queued (all hashes match), `tool_index_ready = True` set immediately without waiting for background task.

**What to watch in logs**
```
# Good
"tool_index_ready = True"
"RAG tool retrieval"

# Bad
"RAG tool retrieval search failed; falling back"   ← pgvector issue
"No routing target for"                            ← routing map broken
"Store is None"                                    ← store not wired up
"PydanticInvalidForJsonSchema"                     ← schema extraction bug still present
```

1Jamie:
I implemented a ```safe_convert``` utility in ```helpers.py``` that wraps ```voluptuous_openapi.convert``` with a more capable serialization logic. (after clearing my index and fixing the lack of the catch in the tuple for that this issue was revealed underneath)
Enum Extraction now detects ```SelectSelector``` and extracts its options into an OpenAPI enum to make sure the LLM can only spit out something valid in that case too. Should also deal with any failed ```hash()``` checks in ```voluptuous_openapi```  causing the system to crash by defaulting them to a string instead.  Addresses the issue in both places for the tuple and this newly found crash. Also added a test ```test_unhashable_selector.py``` for this exact scenario to make sure any unhashable situations are tested against going forward. 

also addressed a ```TypeError``` crash in the RAG tool retrieval logic. There was an issue caused by tool_retrieval_limit being used without explicitly checking its type and making it the right one for the situation. just a simple handling of if it has decimal or not.

though for my own forgetful sanity i did combine all the tests into a local bash script for myself to verify everything before commits to hopefully catch the CI pieces i might forget  again in the future.... i did the typecheck before the last commit so that was my own messup of the order of operations. i had gone through and then in a different commit i did the fix of the noqa's and forgot to check after that with the typecheck so apologies. 

make lint: all checks pass
make typecheck: 0 errors
make test: all 714 tests rul and pass
walked through listed scenarios, tool attachment actuates when its supposed to, fails to full set when its supposed to and follows the assigned parameters. 

to prevent any future cold start  issues from falling through the cracks again i cleared out my postgres tables and grabbed a snapshot in proxmox that i can jump to for a fresh testing setup every time to test the full lifecycle.

though i will apologies i have made you put in this much leg work on testing my stuff and chasing out the problems though it is greatly appreciated! at least its for a sizable addition? lol, python i would not call my best language, but fingers crossed i got it this time! i went through it with a fine tooth comb ran all the tests, scenarios and verified everything in my test env was sanitized before hand with the only complaint anything spit out at me being the single warning in the test suite for pydantic v1 but thats unrelated to these changes.  (oh and i removed my debug files, sorry!!! i am a mess, thanks for dealing with me!)

goruck:
Closing in favor of a new PR from goruck:upstream-sync so the source branch correctly reflects the current maintainer. All commits and 1Jamie's co-authorship are preserved. See replacement PR for continuation.

PR #421, opened 2026-05-20
feat: configurable STT hallucination filter before LLM

alex-mextner:
## Summary
Adds a **user-configurable** pre-LLM filter that silently drops phantom STT transcriptions produced from silence or background noise (e.g. fan hum, TV audio). This prevents wasted tokens and unwanted TTS responses when the microphone accidentally picks up non-speech sounds.

## How it works
- A new **global option** `stt_hallucination_patterns` lets users define substrings to ignore.
- The option defaults to **empty** — no hardcoded patterns. The user decides what to filter.
- Before invoking LangGraph / LLM, `_async_handle_message_active()` checks whether the raw STT text contains any user-defined pattern.
- If matched, the conversation returns an empty `ConversationResult` immediately — no LLM call, no tool execution, no TTS.

## UI
```
Home Generative Agent → Configure (⚙️)
├─ System prompt
├─ LLM APIs
├─ ...
└─ STT hallucination patterns
   [One pattern per line. Matching is case-insensitive and]
   [partial: 'sub' matches 'Subtitles'. Leave empty to   ]
   [disable the filter entirely.                           ]
```

- **Multiline textarea** — type one pattern per line (no need for commas).
- **Description helper** explains matching rules directly under the field.

## Tests
- `test_conversation_units.py` — 9 unit tests covering:
  - empty / default behaviour
  - comma-separated parsing
  - **multiline (textarea) parsing**
  - **mixed delimiters**
  - whitespace trimming
  - match / no-match logic (case-insensitive, partial)

## Checklist
- Syntax verified via `py_compile` ✅
- Unit tests pass (9/9 new, 5 old async tests skipped due to missing pytest-asyncio in HA container) ✅


goruck:
## Code Review — gstack /review

### Scope note

The PR title describes one feature but the diff contains three independent features:

1. **STT hallucination filter** (stated) — \`conversation.py\`: pre-LLM filter that drops phantom transcriptions matching user-configurable substring or exact-match patterns before they reach the agent graph.
2. **LLM provider fallback chains** (unstated) — \`core/fallback.py\`: new \`FallbackChatModel\`, \`FallbackVLM\`, \`FallbackEmbeddings\` wrappers with a \`CircuitBreaker\` that transparently retries on the next provider when the primary fails.
3. **Fallback provider UI** (unstated) — \`flows/feature_subentry_flow.py\` + \`core/subentry_resolver.py\`: multi-select "Fallback providers (ordered)" field in the feature subentry config flow; \`resolve_fallback_chains()\` builds the ordered provider lists consumed by \`__init__.py\`.

Features 2 and 3 are tightly coupled, but Feature 1 is completely independent (different file, different concept, no shared code). Worth splitting next time.

---

### Critical issues — please fix before merging

**1. \`asyncio.run()\` crash in sync embedding shims (\`core/fallback.py\`)**

\`FallbackEmbeddings.embed_documents\` and \`embed_query\` use \`asyncio.run()\` to bridge sync callers into the async chain. This raises \`RuntimeError: This event loop is already running\` in HA's async environment — these methods will crash every time they're called. Replace with \`NotImplementedError\` since sync embedding is not a supported path here:

```python
def embed_documents(self, texts: list[str]) -> list[list[float]]:
    msg = "Use aembed_documents in async context"
    raise NotImplementedError(msg)

def embed_query(self, text: str) -> list[float]:
    msg = "Use aembed_query in async context"
    raise NotImplementedError(msg)
```

**2. API key exposure in logs (\`core/fallback.py\`)**

Raw exception objects are passed directly to \`LOGGER.warning(..., err)\`. Provider HTTP responses (OpenAI, Anthropic, Gemini) routinely echo the API key or bearer token in their error bodies. Truncate before logging:

```python
def _safe_err_summary(err: Exception) -> str:
    return f"{type(err).__name__}: {str(err)[:200]}"
```

Apply to all \`LOGGER.warning\` calls and \`HomeAssistantError\` messages in the file.

**3. Shared \`CircuitBreaker\` across all model types (\`__init__.py\`)**

A single \`CircuitBreaker\` instance is shared across chat, VLM, summarization, and embedding models. A rate-limit burst on Sentinel's VLM calls will disable the main conversation agent. Use four independent circuit breakers: \`_chat_cb\`, \`_vlm_cb\`, \`_sum_cb\`, \`_emb_cb\`.

**4. Fallback chain wiring drops model config (\`__init__.py\`)**

The fallback chain construction re-instantiates raw provider instances for the primary model, silently discarding the already-configured \`model_name\`, \`temperature\`, and other per-feature settings. The pre-configured primary model instance should be used directly; only providers 2..N should be added as fresh fallback instances.

---

### Greptile bot comment — false positive

The Greptile comment references an old commit (\`07c4b7f962\`) that contained \`_is_tv_noise()\` with hardcoded patterns. The current code uses \`_is_stt_hallucination()\` with fully user-configurable patterns. The concern is already resolved.

---

### Informational (not blocking)

- **\`CONF_FALLBACK_ALERT_ENABLED\` is dead code** (\`const.py\`): defined but never read anywhere; either wire it up or remove it.
- **Zero test coverage for \`core/fallback.py\`** (397 lines — \`CircuitBreaker\`, all three \`Fallback*\` classes). Worth adding at minimum: CB threshold/cooldown behavior, \`ainvoke\` fallback-on-retryable-error, \`ainvoke\` raise-on-non-retryable-error, circuit-broken provider skipped.
- **Zero test coverage for \`resolve_fallback_chains\`**.
- **DRY**: the 5-line retry loop (\`circuit_breaker.is_available → try/except → record_failure → alert → continue\`) is duplicated across 5 methods in 3 classes. A private \`_run_with_fallback(chain, fn)\` coroutine would remove the repetition.
- **\`FallbackChatModel.astream\` mid-stream fallback**: if the primary provider fails after yielding partial chunks, the fallback starts from the beginning — the caller receives a truncated first response followed by a full second response concatenated together. Worth documenting or buffering.
- **\`generate_embeddings\` in \`core/utils.py\`**: uses \`isinstance(emb, GoogleGenerativeAIEmbeddings)\` to choose vector dimensions. When \`FallbackEmbeddings\` wraps a Gemini provider this check fails and wrong dimensions would be used. Needs to inspect the chain's first provider instead.
- **Non-deterministic fallback ordering**: \`resolve_fallback_chains\` iterates \`dict.values()\` — insertion order in Python dicts is stable within a run but provider subentry registration order can vary between HA restarts. The primary provider could silently change. Consider an explicit \`priority\` field or sort by \`entry_id\` for stable ordering.
- **STT text logged at DEBUG**: the raw transcription is logged before the hallucination check, which could log PII (speech content). Consider eliding or hashing in the log line.

goruck:
Thanks @alex-mextner  for continuing to iterate on this PR and pushing the latest STT filter UI commits. I pulled those changes in and merged them with the local review fixes.

What changed in this update:
- Kept the STT filter configuration in its own bottom section, but switched the two filter inputs from chip/select controls to multiline text fields. I think the chips are confusing once real strings are added, especially for phrase-like filters.
- Moved the critical-action PIN entry directly under the Require critical action PIN switch so related controls stay together.
- Added missing fallback_provider_ids labels and descriptions for provider submenus in the translated strings.
- Fixed fallback provider setup so chat, VLM, and summarization fallback models are built with the configured provider/category settings instead of raw provider defaults.
- Fixed embedding fallback setup so embedding fallback chains use embedding clients, skip unavailable primary embedding clients, and log when an embedding fallback provider actually activates.
- Added regression coverage for fallback wiring, embedding fallback activation logging, and options schema ordering/control types.

Verification completed on the updated branch:
- ./hga/bin/pytest tests/custom_components/home_generative_agent/test_options_schema.py tests/custom_components/home_generative_agent/test_conversation_units.py tests/custom_components/home_generative_agent/test_fallback.py tests/custom_components/home_generative_agent/test_fallback_setup.py -> 44 passed
- hga/bin/ruff check custom_components/home_generative_agent/__init__.py custom_components/home_generative_agent/config_flow.py custom_components/home_generative_agent/core/fallback.py tests/custom_components/home_generative_agent/test_options_schema.py tests/custom_components/home_generative_agent/test_fallback.py tests/custom_components/home_generative_agent/test_fallback_setup.py -> passed
- hga/bin/pyright -> 0 errors, 0 warnings
- JSON validation for strings.json and en/tr/ru translations -> passed
- git diff --check origin/main...HEAD -> passed

Thanks again for the concurrent work here. I kept the intent of the STT section iteration, but resolved it toward multiline inputs so the resulting options flow should be clearer for real-world filter values.

goruck:
Thanks Alex for the continued work on this PR. I think this is ready for your review again.

Since the last summary comment, I pushed a set of follow-up changes based on integration testing:

- Added clearer tool retrieval diagnostics so fallback logs distinguish index-not-ready, index-failed, and vector-search-returned-no-usable-candidates cases.
- Fixed setup-time fallback when the primary provider is unavailable after restart. The chain now skips unavailable placeholder models instead of letting `NullChat` return `LLM unavailable.` and short-circuit the configured fallback.
- Updated effective deployment metadata after setup-time fallback selection. If chat falls back from an edge provider to a cloud provider, the runtime deployment metadata now reflects cloud, so chat-priority/session logs are no longer misleading.
- Added setup-time fallback logs that explicitly say which provider was selected and that the selection remains active until integration reload or Home Assistant restart.
- Reduced log noise for categories with no configured fallback. Unconfigured VLM/summarization fallback now logs at debug level, while configured-but-unusable fallback chains still warn.
- Updated `docs/configuration.md` to document provider fallback behavior, including per-feature scope, setup-time versus runtime fallback, sticky-until-reload behavior, and model default resolution.
- Documented that fallback model settings come from the fallback provider subentry first (`chat_model`, `vlm_model`, `summarization_model`), then fall back to `const.py` / `MODEL_CATEGORY_SPECS` recommended defaults.
- Added regression coverage for fallback setup wiring, unavailable primary providers, embedding fallback setup, and the options/schema changes from the earlier review work.

Integration testing now shows:

- Chat setup fallback selecting the configured cloud provider when Ollama is unavailable.
- VLM setup fallback selecting the configured Anthropic/cloud provider.
- A camera prompt successfully calling `get_and_analyze_camera_image`, using the VLM fallback, and returning the result through the chat model.
- Tool RAG retrieving the expected camera tools once the index is ready.

Focused verification run locally included fallback setup/wrapper tests, tool retrieval/conversation tests, Ruff, Pyright, formatting checks, JSON validation, and `git diff --check` across the touched changes.

Thanks again for the PR and for iterating on the provider/fallback UI. Could you review the latest branch state when you get a chance?

alex-mextner:
Thank you! Going to see today

goruck:
Thanks again @alex-mextner for the original STT hallucination filter and model-provider fallback work in this PR. I also added explicit attribution for your contribution in README.md and CHANGELOG.md so it is visible outside GitHub's contributor graph.

PR #460, opened 2026-07-06
fix: auto-resolve Ring-MQTT and Reolink motion sensors to camera entities

### LINKED ISSUE #459: Documentation gap: "Proactive motion-triggered analysis" requires manual automation setup in Advanced mode
## System Information

version | core-2026.7.1
-- | --
installation_type | Home Assistant OS
dev | false
hassio | true
docker | true
container_arch | aarch64
user | root
virtualenv | false
python_version | 3.14.6
os_name | Linux
os_version | 6.18.37-haos
arch | aarch64
timezone | Europe/London
config_dir | /config

<details><summary>Home Assistant Cloud</summary>

logged_in | true
-- | --
subscription_expiration | 1 January 2018 at 00:00
relayer_connected | false
relayer_region | null
remote_enabled | true
remote_connected | false
remote_server | null
alexa_enabled | true
google_enabled | false
cloud_ice_servers_enabled | true
certificate_status | null
instance_id | cf8cf4de8dce44c8a63ec9c8e6fe4095
iot_state | disconnected
iot_tries | 0
can_reach_cert_server | ok
can_reach_cloud_auth | ok
can_reach_cloud | ok

</details>

<details><summary>HACS</summary>

GitHub API | ok
-- | --
GitHub Content | ok
GitHub Web | ok
HACS Data | ok
GitHub API Calls Remaining | 4996
Installed Version | 2.0.5
Stage | running
Available Repositories | 3510
Downloaded Repositories | 10

</details>

<details><summary>Home Assistant Supervisor</summary>

host_os | Home Assistant OS 18.1
-- | --
update_channel | stable
supervisor_version | supervisor-2026.06.2
agent_version | 1.9.0
docker_version | 29.5.3
disk_total | 56.6 GB
disk_used | 14.7 GB
nameservers | 192.168.1.1, 2606:4700:4700::1001, 2620:fe::fe, 8.8.4.4, 1.0.0.1, 2620:fe::9, 1.1.1.1, 8.8.8.8, 2606:4700:4700::1111
healthy | true
supported | true
host_connectivity | true
supervisor_connectivity | true
ntp_synchronized | true
virtualization | 
board | odroid-n2
disk_life_time | 0 %
supervisor_api | ok
version_api | ok
installed_addons | Mosquitto broker (7.1.0), Zigbee2MQTT (2.12.1-1), Ring-MQTT with Video Streaming (5.9.3), Matter Server (9.0.4), File editor (6.0.0), Everything Presence Zone Configurator (2.2.4), Terminal & SSH (10.3.0), OpenThread Border Router (3.0.1), PostgreSQL with pgvector (0.1.0), Home Assistant MCP Server (7.9.0)

</details>

<details><summary>Dashboards</summary>

dashboards | 6
-- | --
resources | 4
views | 22
mode | storage

</details>

<details><summary>Network Configuration</summary>

adapters | lo (disabled), end0 (enabled, default, auto), hassio (disabled), docker0 (disabled), vethe662a39 (disabled), veth54fec33 (disabled), veth97ac4e6 (disabled), veth27c9b87 (disabled), veth881773f (disabled), vethf394be3 (disabled), veth557b7fa (disabled), vethb9d7f93 (disabled), wpan0 (disabled), vethc9e64fa (disabled), veth3daeffc (disabled), veth347796e (disabled)
-- | --
ipv4_addresses | lo (127.0.0.1/8), end0 (192.168.1.249/24), hassio (172.30.32.1/23), docker0 (172.30.232.1/23), vethe662a39 (), veth54fec33 (), veth97ac4e6 (), veth27c9b87 (), veth881773f (), vethf394be3 (), veth557b7fa (), vethb9d7f93 (), wpan0 (), vethc9e64fa (), veth3daeffc (), veth347796e ()
ipv6_addresses | lo (::1/128), end0 (fd32:f126:6bfb:c279:98bb:422:4c0:d2/64, fe80::349:af1:a569:b121/64), hassio (fd0c:ac1e:2100::1/48, fe80::9038:81ff:fe7c:c707/64), docker0 (fd22:2733:2f41::1/64, fe80::ec40:c4ff:fe80:f807/64), vethe662a39 (fe80::a4e6:15ff:fe4a:af08/64), veth54fec33 (fe80::34c1:a9ff:fe48:9512/64), veth97ac4e6 (fe80::3c12:42ff:feaf:b4cf/64), veth27c9b87 (fe80::7000:eff:fe41:a197/64), veth881773f (fe80::c83a:81ff:fec9:6aa6/64), vethf394be3 (fe80::50b0:e1ff:fed2:5004/64), veth557b7fa (fe80::f4fa:80ff:fed3:f7a5/64), vethb9d7f93 (fe80::8b:2ff:fe65:23a6/64), wpan0 (fd81:892f:98be:4859:0:ff:fe00:fc11/64, fdac:f698:a63b:1:e575:40f4:a0cd:8209/64, fd81:892f:98be:4859:0:ff:fe00:fc38/64, fd81:892f:98be:4859:0:ff:fe00:fc10/64, fd81:892f:98be:4859:0:ff:fe00:fc00/64, fd81:892f:98be:4859:0:ff:fe00:8800/64, fd81:892f:98be:4859:1c28:7ea8:634d:9c9e/64, fe80::e008:c372:58e:472e/64), vethc9e64fa (fe80::1c53:fbff:fe67:a0d6/64), veth3daeffc (fe80::68b7:8bff:fed6:28b8/64), veth347796e (fe80::1cc6:6cff:fe37:6755/64)
announce_addresses | 192.168.1.249, fd32:f126:6bfb:c279:98bb:422:4c0:d2, fe80::349:af1:a569:b121

</details>

<details><summary>Recorder</summary>

oldest_recorder_run | 21 June 2026 at 03:03
-- | --
current_recorder_run | 6 July 2026 at 05:16
estimated_db_size | 907.41 MiB
database_engine | sqlite
database_version | 3.53.2

</details>

### Checklist

- [x] I have enabled debug logging for my installation.
- [x] I have filled out the issue template to the best of my ability.
- [x] This issue only contains 1 issue (if you have multiple issues, open one issue for each issue).
- [x] This issue is not a duplicate issue of any [previous issues](https://github.com/ludeeus/integration_blueprint/issues?q=is%3Aissue+label%3A%22Bug%22+)..

## Documentation gap: "Proactive motion-triggered analysis" requires manual automation setup in Advanced mode

**Summary**

The README and integration description list "Proactive motion-triggered analysis with anomaly detection" as a core feature. After setting up HGA via **Advanced** mode (configuring Model Provider, Conversation, Camera Image Analysis, and Sentinel as individual subentries across several sessions, rather than using Basic setup), I found that no automation actually connects a camera's motion sensor to `save_and_analyze_snapshot`. The service exists and works correctly when called manually, but nothing calls it automatically.

**How I found this**

I confirmed via a real-world test: 15 genuine Ring camera motion events over a 2.5-hour window (car being moved and washed in the driveway, two people walking in and out of frame) produced **zero** Sentinel audit records of any kind — not just zero for camera-specific rules like `vehicle_detected_near_camera_home`, but zero activity logged at all for that camera during that window. This fully explained why camera-based Sentinel rules (`vehicle_detected_near_camera_home`, `unknown_person_camera_*`, `pet_detected_at_night_no_occupancy`, `camera_missing_snapshot_night_home`) had never fired on organic activity in my setup — the underlying vision analysis simply never ran unless I called the service manually.

**Suspected cause**

Looking at the setup flow documentation, Basic mode is described as enabling "all features with recommended defaults." It's possible Basic mode auto-creates the motion→analysis wiring as part of those defaults, while Advanced mode (which I used, configuring each feature individually) leaves this connection for the user to build themselves. I haven't confirmed this against the source, but the setup-mode difference is the most plausible explanation I could find.

**Suggested fix**

Regardless of whether this is Basic-vs-Advanced-specific: the documentation should explicitly state, wherever "proactive motion-triggered analysis" is described as a feature, that **Advanced-mode users (or all users, if this turns out to apply universally) need to create one automation per camera** triggering `home_generative_agent.save_and_analyze_snapshot` on motion. A short example automation in the docs (motion sensor → snapshot service call, with a suggested cooldown to avoid hammering the vision model during sustained activity) would have saved a fair amount of debugging time.

**What I built as a workaround**, for reference, in case it's useful as a documentation example:

```yaml
alias: HGA - <Camera Name> - motion analysis
triggers:
  - trigger: state
    entity_id: binary_sensor.<camera>_motion
    to: "on"
conditions:
  - condition: template
    value_template: >
      {{ (as_timestamp(now()) - as_timestamp(this.attributes.last_triggered
      | default('1970-01-01 00:00:00'))) > 600 }}
actions:
  - action: home_generative_agent.save_and_analyze_snapshot
    target:
      entity_id: camera.<camera>_snapshot
mode: single
```

(10-minute cooldown via a template condition on the automation's own `last_triggered`, since there's no native HA construct for "time since this automation last fired.")

Happy to provide more detail on my setup/config if useful for reproducing or documenting this.

### Reproduction steps

1. As above -  Documentation / Code gap
2.
3.
...

### Debug logs

```text
1.As above - Documentation / Code gap - therefore no functionality to debug
```

### Diagnostics dump

_No response_

goruck:
## Summary

Fixes Ring-MQTT (and similar) cameras not triggering proactive video analysis due to incorrect motion sensor → camera binding.

**Root cause:** `_resolve_camera_from_motion` used name heuristics that matched interior Ring Alarm PIR sensors (e.g. `binary_sensor.front_door_motion`) to the doorbell camera instead of the doorbell's own motion sensor (`binary_sensor.front_door_motion_3`). The PIR stays ON for 3+ minutes at 3s snapshot intervals → up to 62 phantom notifications per event.

**Fixes:**
- Introduces a **three-tier motion → camera resolution** in `_resolve_camera_from_motion`:
  1. Explicit override map (UI-configurable in Global Options)
  2. HA device registry lookup — finds the camera entity on the same device as the motion sensor, correctly separating Ring Alarm PIRs from doorbell cameras
  3. Name heuristics (direct substitution, VMD suffix strip, `_motion` suffix strip) — fallback for integrations without device links

- Adds **UI-configurable motion sensor override map** (`CONF_VIDEO_ANALYZER_MOTION_CAMERA_MAP`) in Global Options so users can pin specific sensors to specific cameras without code changes

- Adds **event-driven recording state loop** for cameras that enter a `recording` HA state (UniFi Protect, etc.) with motion loop promotion — if a motion binary sensor fires while a recording loop is already running, the recording loop is cancelled and the motion loop takes ownership of the queue flush

- Adds **UI toggle for perceptual hash (dHash) frame deduplication** (`CONF_VIDEO_ANALYZER_UNIQUENESS_ENABLED`), replacing the hardcoded `_UNIQUENESS_ENABLED = False` constant

- Rewrites `docs/camera-entities.md` Proactive Video Analysis section with per-camera-type documentation (Axis VMD, ring-mqtt, recording-state cameras, official Ring HA integration), three-tier resolution documentation, VLM quality requirements, notification modes, and advanced options

## Test plan

- [x] 8 new unit tests: all three resolution tiers, motion interval sanity check, recording poll mutual exclusion — all pass
- [x] Full suite: 1306 passed
- [x] Deployed to HA and tested with live Ring-MQTT cameras (verified correct sensor binding via device registry)

Closes #459

## Credit

Special thanks to **@andymcmanus** for detailed bug reports, debug log analysis, and patient re-testing across multiple fix iterations. Their investigation — including capturing the exact sensor state timelines and identifying the interior PIR vs. doorbell sensor collision — was essential to diagnosing the true root cause and validating the device-registry fix.


andymcmanus:
Thanks for the quick turnaround on this — tested PR #460 today. Results below, good and bad.

**Setup process**

To make sure I had a clean test, I copied the HGA directory to a `.bak` folder first. That actually broke HA on reboot — errors referencing the `.bak` naming, so I learned the hard way that a backup can't live alongside the original in the same custom_components structure. I moved the backup fully out to `/tmp`, confirmed HA reloaded cleanly with just the original files back in place, then copied over the two changed files from the PR (`video_analyzer.py` and the manifest). Rebooted again — manifest showed the version bump, HGA loaded with no errors. Clean single-copy install confirmed before testing.

**I had also manually disabled the 5 workaround automations I'd built** (motion sensor → `save_and_analyze_snapshot`, one per camera) *before* this reboot — so they were off before the patched code ever ran. Ruling those out as a contributing factor.

**What happened**

Walked out my front door once to trigger Ring motion — a single real-world event. Got hit with a burst of roughly 100 phone notifications in quick succession, a brief pause, then another burst started. I disabled the HGA integration and rebooted HA to stop it, which was necessary but means I don't have HA logs from during the incident — the reboot started a fresh log and I lost the exact trace.

**A guess, not a confirmed diagnosis**

No hard evidence for this, so take it as a lead rather than a diagnosis: from your Medium writeup on the Video Analyzer architecture, snapshots go through per-camera asyncio queues with perceptual-hash/heartbeat dedup. The burst→pause→burst pattern feels consistent with the new camera-resolution fallback (trying `camera.<base>` then `camera.<base>_snapshot`) possibly enqueuing more than one candidate entity per motion event rather than stopping at the first match, with the queue draining (burst), then something (heartbeat cycle?) triggering a refill (second burst). Could easily be wrong — just the most plausible mechanism I could think of given the symptom shape.

**Current status on my end**

I've reverted back to the original `video_analyzer.py` (kept a full backup of the pre-patch state in `/tmp` throughout, so the revert was clean) and re-enabled my 5 workaround automations, which were never implicated in this and have been stable. Back to a known-good state.

Happy to test again if you want me to try a specific debug-logging config, or if a revised version of the branch addresses this — just let me know what would be most useful. Didn't want to keep hammering my own phone with 100-alert bursts trying to get more detail without some guidance on what to capture first.

goruck:
Thanks for testing so thoroughly and for the detailed write-up — the burst→pause→burst description was exactly what we needed to find the second bug.

**What caused the notification flood**

The camera resolution fix worked correctly — your Ring motion sensor was being paired with the right camera entity for the first time. The flood was caused by a separate pre-existing issue that only became visible once Ring cameras started working: the motion snapshot loop was running at the same 1.5-second interval used for always-on recording cameras. Ring-MQTT holds `binary_sensor.*_motion` in the `on` state for the full recording duration (60–180 seconds by default in the Ring app), so at 1.5s intervals a single motion event produces 40–120 snapshots. With caption deduplication imperfect on dynamic outdoor scenes, most of those generated a notification.

**The fix**

A second commit on this branch introduces `VIDEO_ANALYZER_MOTION_SCAN_INTERVAL = 8` seconds, used only by the motion-triggered snapshot loop. The 1.5s recording-camera polling is unchanged. At 8s, a 60-second Ring motion window produces ~8 snapshots rather than 40, and a 3-minute window produces ~22 — a much more manageable number for deduplication to handle.

**Would you be willing to test again?**

Before testing, it's worth checking your Ring app's motion recording duration setting so you know roughly how long your motion windows are — that'll help calibrate expectations. The `image.<cam>_last_event` entity updating on motion is still the cleanest first check; Sentinel audit records are the confirmation that the full pipeline is working.

Apologies for the extra round-trip, and thanks again for sticking with it.

andymcmanus:
Tested the `VIDEO_ANALYZER_MOTION_SCAN_INTERVAL` fix today. The camera resolution piece is clearly working — but there's a new, more serious issue.

**What happened**

Single real motion trigger at the front door. Instead of a manageable ~8-22 snapshots, I got a burst of phone notifications at roughly 5-second intervals that didn't stop on its own for about 4 minutes — 48 notifications total, all seemingly the same or near-identical content, before it finally stopped.

**Ruling out the obvious explanations**

- My own workaround automations (motion sensor → `save_and_analyze_snapshot`, one per camera) were manually disabled before this test, so they weren't a contributing factor.
- I checked the motion sensor's actual state partway through the storm: it had already returned to `off`. The notifications kept arriving well after the real-world motion had stopped — so this isn't "lots of legitimate motion," it's something looping independently of the sensor.
- Motion recording duration in the Ring app is in the 60-180s range as you'd expect — nothing unusual there that would explain a 4-minute, 48-notification event on a single trigger.

**A clue that might help narrow it down**

I pulled the last 100 records from `get_audit_records` afterward — the storm doesn't appear in there at all. The most recent entry is a routine, pre-existing finding type from earlier in the day; none of the 48 rapid notifications show up as Sentinel findings. That suggests whatever looped was **not** going through the normal finding → audit → notify pipeline (which has its own cooldown/dedup protections) — it seems to be a separate, more direct notification path, possibly inside the video_analyzer's own anomaly-detection notification logic rather than Sentinel's rule engine.

**Status on my end**

I disabled HGA once the storm stopped (it did stop on its own — I didn't need to intervene, though I was ready to). I'm hesitant to trigger this again given the notification volume it produces, but happy to test a further revision, or to try again with debug logging enabled first if that would help you localize it — just let me know what's most useful and I'll set it up carefully before testing.

**Additional finding — precise timing evidence from disk, and a separate silent-failure bug**

A few hours after the storm, I found the actual snapshot files on disk for that event: `/media/snapshots/camera_front_door_snapshot/`. 27 files, timestamped from `17:07:04` to `17:10:38` — **exactly 214 seconds (3m34s) of continuous capture at a steady ~8-second interval**, all from a single brief front-door event. So the new interval fix is working correctly on its own terms (8s spacing, confirmed), but the loop clearly isn't exiting when it should — it ran for well over 3 minutes on an event that should have lasted well under a minute. Notably, I checked the motion sensor's live state partway through the original storm and it had already returned to `off` — so whatever's keeping the loop running doesn't appear to be tied to the real-time state of `binary_sensor.<cam>_motion`. That's probably the more precise way to describe this bug: not "the interval is too short" (already fixed) but "the loop's exit condition isn't correctly tied to the motion sensor actually returning to idle."

One number worth flagging as a discrepancy rather than a tidy match: ~48 phone notifications during the storm, but only 27 snapshot files. More notifications than images — so either some snapshots produced multiple separate notifications, or these are two related but distinct symptoms of the same underlying loop rather than a 1:1 relationship. Passing it along as-is rather than forcing it to reconcile.

Separately — later in the day, with HGA re-enabled and otherwise apparently healthy, a single new front-door event produced no notification and no Sentinel finding at all. The workaround automation fired correctly (confirmed via its trace), but the log showed: `[camera.front_door_snapshot] Snapshot failed to appear on disk after waiting: /media/snapshots/camera_front_door_snapshot/snapshot_20260707_170646.jpg` (WARNING level, no accompanying error). This happened ~52 seconds after HGA's own startup/health-check log lines for that reload, so it's plausibly a cold-start race condition (something not fully ready yet right after reload) rather than a persistent problem — but wanted to flag it as a third, distinct issue in case it's a known or easy one to address alongside the others.
 (updated with new findings 18:08 BST)

goruck:
One more update on the branch before you test again.

**What changed**

The previous fix (8-second scan interval) would have reduced the notification count but at the cost of potentially missing brief events between snapshots. Instead the motion snapshot loop has been redesigned as a fixed burst: on the rising edge of the motion sensor, HGA takes 3 snapshots spaced 3 seconds apart, then stops — regardless of how long the sensor stays `on`.

This means a Ring-MQTT sensor held `on` for 120 seconds produces exactly the same 3-frame analysis window as a brief-pulse sensor. No tuning of scan intervals or similarity thresholds needed; the behaviour is the same across all camera integrations.

The branch (`fix/ring-mqtt-motion-resolution`) now has all three fixes: camera resolution, the old scan-interval change (now replaced), and this burst approach. Happy to hear how it goes whenever you get a chance to test.

andymcmanus:
Tested the burst-based redesign today. Good news and a much clearer picture of the actual root cause than my last report — I think I mischaracterized this a bit last time.

The burst fix itself is working correctly

Single motion event this time produced 3 notifications, not 27 — the runaway loop is genuinely gone. Progress.

But there are two completely separate notification paths, and only one of them is deduplicated

Screenshot of the 3 notifications received from a single walk-out event (attached — happy to send the raw file if useful):


"Home Update — Outdoor activity while alarm disarmed": "Front Door Snapshot detected unrecognized outdoor activity 1 min ago. The alarm has been disarmed since 22 Jun at 8:06 AM. Arm the alarm or view th..."
"Camera Alert from front_door_...": "The image shows a front porch with a brick wall and white ceiling. A potted plant is on the porch, and a number '01' is visible in the backg..."
"Camera Alert from front_door_...": "The image shows a brick house with a white porch and a black fence, set against a cloudy sky."


Notification #1 is a proper Sentinel finding — title format, phrasing, and content all match what I'd expect from the normal finding/audit pipeline (I've confirmed this type does show up correctly in get_audit_records).

Notifications #2 and #3 are a different notification path entirely — different title format ("Camera Alert from X" vs. the Sentinel-style title), and critically: these never appear in get_audit_records at all. I checked this specifically after the original storm — 100 recent audit records pulled, none of them corresponded to the storm, even though ~48 notifications had gone to my phone. I think this "Camera Alert" path is the raw, direct video_analyzer notification described in your Medium writeup (fired per-snapshot, independent of Sentinel's rule engine), and unlike the Sentinel path, it doesn't appear to have working deduplication.

This would explain both incidents with one root cause


The original storm: every one of those ~48 notifications was very likely this same "Camera Alert" path firing once per snapshot in the (at the time, unbounded) motion loop, completely bypassing whatever similarity/dedup logic exists for Sentinel findings.
Today's test: 2 near-duplicate "Camera Alert" notifications from what should have been 3 fused-into-one snapshots, worded differently enough (porch/plant/house-number vs. house/fence/sky — same physical scene) that if there is a similarity check on this path, it isn't catching this pair. I'm currently running moondream as the vision model (not qwen3-vl, for unrelated hardware reasons — an unsupported iGPU on my end, not a fault of HGA), and moondream is a much smaller/more basic model than qwen3-vl. Given your prompt engineering explicitly relies on "consistent phrasing across similar frames" to make the dedup/similarity logic work, it's plausible moondream just doesn't follow that instruction as reliably as qwen3-vl does — worth keeping in mind if others report this and are also running a lighter-weight VLM.


One more small residual item

Only 2 snapshot files actually appeared on disk for this 3-snapshot burst (/media/snapshots/camera_front_door_snapshot/), matching the same kind of silent snapshot-write issue I flagged before (Snapshot failed to appear on disk after waiting, WARNING-level, no error). Feels like a separate, lower-priority reliability issue rather than connected to the notification duplication above.

Suggested next step

Might be worth checking whether the "Camera Alert from X" path has its own similarity/dedup check at all, separate from the one governing Sentinel findings — if it's missing entirely (rather than just failing due to model-specific phrasing variance), that'd explain the original storm more directly than a threshold tuning issue would.

Happy to keep testing — this feels like real progress, just a different bug underneath than the one we were chasing before.

andymcmanus:
A few more findings since my last update — one new, separate bug, and a deeper look at the phantom notification bursts.

**New bug: Sentinel Discovery references a model that isn't installed**

Disabling HGA this evening surfaced this traceback (only visible because the integration unload tried to gracefully stop the Discovery Engine's background task mid-call):

```
File ".../sentinel/discovery_engine.py", line 328, in _run_once
    result = await run_sentinel_model_call(...)
...
ollama._types.ResponseError: model 'gpt-oss' not found (status code: 404)
```

I enabled `sentinel_discovery_enabled` a few days ago. It looks like Discovery has been silently failing on every cycle since then, trying to call a model (`gpt-oss`) that was never pulled on my Ollama instance (I only have `qwen3:8b` and `moondream:latest` installed). Might be worth Discovery falling back to whatever model is actually configured for Sentinel's other LLM calls, rather than a hardcoded default that assumes a specific model is present.

**Confirmed: the snapshot-write failure from my last report is recurring, not a one-off**

Same warning, different timestamp, happening independently of the notification storm:
```
[camera.front_door_snapshot] Snapshot failed to appear on disk after waiting: snapshot_20260707_183550.jpg
```

**The phantom notification bursts — a new angle, and I think this is the key finding**

I had a second, smaller incident today: 6 "Camera Alert from front_door_..." notifications with no corresponding Ring motion event at all (nothing in the Ring app's activity log for that window). Same pattern as the original storm — just smaller.

This time I checked thoroughly before disabling anything: **zero audit records** for that window (confirmed via `get_audit_records`), and **zero system log entries at any level** (WARNING through ERROR, unfiltered, across the full window) — nothing at all between a routine reload at 17:36 and the notifications firing at ~19:33-19:37. I also went back and confirmed the same is true for the original 48-notification storm from a few days ago.

I think this means the code path generating these "Camera Alert" notifications has no logging instrumentation around it at all — which would explain why neither of us can find anything, regardless of how hard we search. If that's right, the most useful thing on my end would be for you to point me at whichever function actually sends these (I'm assuming it's separate from the Sentinel finding/audit pipeline, based on the title format and total absence from `get_audit_records`) so I can either add temporary logging myself before the next test, or you could add some in a debug branch — whichever's easier for you. Right now I genuinely have no way to catch this in the act beyond "it happened, here's a screenshot," which isn't giving you much to work with.

Let me know what would help most.

andymcmanus:
Separate question while we're at it, not related to the camera bugs — I've been setting up face recognition (`face_api_url`) and wanted to check I'm not missing something obvious.

I couldn't find any documentation naming what's actually expected on the other end of `face_api_url` — I ended up reverse-engineering the contract from `recognize_faces()` and `validate_face_api_url()` in the source (`GET /status`, `POST /analyze` with a multipart `file` field, expecting `{"faces": [...]}` back). That doesn't match CompreFace's actual API, or Double Take's, or anything else I could find documented for this kind of setup — so I ended up building a small custom FastAPI service around InsightFace to match it.

Is that the expected/intended path — build a minimal service matching this specific contract yourself — or is there a companion repo, reference implementation, or existing tool you'd recommend instead that I've just missed? Happy to be pointed at docs if this is covered somewhere and I didn't find it.

goruck:
Another update — a fair bit has changed based on your findings, so worth explaining the full picture.

**Root cause of the original storm — now identified**

The notification storm wasn't primarily caused by the scan interval or the VLM. It was caused by Ring-MQTT cameras triggering **two separate snapshot paths simultaneously**:

1. The motion sensor (`binary_sensor.*_motion` → `on`) starts the motion snapshot loop
2. Ring cameras also enter `"recording"` state when a clip starts, which causes `_take_snapshots_from_recording_cameras` to pick up the same camera at 1.5s intervals in parallel

Both paths were running concurrently for the full duration of the Ring recording window. The fix adds mutual exclusion: the recording poll now skips any camera already being handled by the motion loop. These two paths are now mutually exclusive.

**The burst approach has been reverted**

The fixed burst (3 snapshots then stop) was too conservative — it would miss meaningful activity that happens after the initial trigger window (someone unloading a car, a package left on the porch, a second person entering the frame). The continuous loop is the right design.

**Motion scan interval is now 3 seconds**

With the dual-trigger fixed, the motion loop is the only path for Ring cameras. The interval has been set to `VIDEO_ANALYZER_MOTION_SCAN_INTERVAL = 3s` (separate from the 1.5s recording poll). 3s captures fast walk-bys reliably while generating half the model calls of 1.5s. A good VLM's semantic dedup handles repetitive frames in long motion windows.

**On your VLM**

One thing worth flagging separately: the 2 near-identical "Camera Alert" notifications you saw from the 3-snapshot burst, and the dedup not catching them, are consistent with moondream generating captions that vary too much between frames of the same scene for the 0.85 similarity threshold to catch. The semantic dedup is designed around VLMs that produce stable, consistent descriptions — qwen3-vl and similar models follow the prompt's phrasing guidance reliably; moondream less so. We'll be adding a note to the docs about minimum VLM quality for the dedup to work as intended. This isn't a criticism of your setup — it's a documentation gap on our side.

**On face recognition**

Saw your question about `face_api_url`. The expected service is [face-service](https://github.com/goruck/face-service) — a companion repo that provides exactly the API HGA expects (`GET /status`, `POST /analyze`). The [installation docs](docs/installation.md) cover setup. Your custom FastAPI/InsightFace wrapper should work too as long as the response schema matches, so nothing broken there — just wanted to point you at the reference implementation in case it's easier.

Happy to hear how the updated branch tests. Thanks again for the thorough reporting — this has been genuinely useful for tracking down a subtle bug.

andymcmanus:
Thanks — the mutual-exclusion explanation fits the motion-triggered storm well, and the burst→pause→burst shape is consistent with two paths sampling the same camera concurrently. Before I re-test, though, I want to flag one case I don't think the root-cause write-up covers, so we don't close this out and have it resurface.

Separate from the storms (which all started from a real motion event), I had a smaller incident with no Ring motion event at all: 6 "Camera Alert from front_door" notifications, nothing in the Ring app's activity log for that window, zero audit records, and zero system-log entries at any level, between a routine reload at 17:36 and the notifications firing ~19:33–19:37.

My concern is that the mutual-exclusion fix keys off the motion loop already owning a camera — but if there's no HA motion event, the motion loop never starts, so "two paths at once" isn't the mechanism in this case. Could a Ring camera entering "recording" on its own (Ring's cloud-side detection, not reflected in binary_sensor.<cam>_motion) still be picked up by _take_snapshots_from_recording_cameras on its own and fire Camera Alerts with no HA motion behind it?

So the specific question: does the mutual-exclusion fix change anything for a camera that enters recording state with no corresponding HA motion event? If not, I suspect the no-motion bursts could survive this fix, and it may point back at the "Camera Alert" path lacking its own dedup rather than at the dual-sampling.

Happy to test the new build regardless — I'll watch specifically for any Camera Alerts that fire without a matching Ring activity entry. And if it'd help, point me at the function that sends these "Camera Alert" notifications and I'll add temporary logging around it before testing, so we can actually catch one in the act — right now that path has no instrumentation, which is why I've only ever been able to give you screenshots.

andymcmanus:
Tested the current branch head (68e5dbc, v3.14.27) today. Camera resolution is solid — every alert resolved to front_door_snapshot correctly. But the notification volume is not fixed, and this run isolates why.

Single front-door event → 42 snapshots → 42 notifications, exactly 1:1.
One real walk-out. 42 snapshot files in /media/snapshots/camera_front_door_snapshot/, timestamped 19:58–20:00 (~3s spacing, matching VIDEO_ANALYZER_MOTION_SCAN_INTERVAL). 42 "Camera Alert" notifications on my phone. A perfect 1:1 — every captured frame produced exactly one notification. Nothing deduplicated: not the mutual-exclusion change, not the new dhash uniqueness gate, not the caption dedup. (My earlier storm was 48 notifications / 27 files — more notifications than files; this one is cleanly 1:1, which points at one-notification-per-frame with zero suppression rather than double-sampling.)

Why the dedup didn't fire — screenshot attached.
The consecutive alerts are all the same static front porch (IR/grayscale, evening), but moondream describes it differently every time:
- "a porch with a brick wall and a white column"
- "a brick house with a porch and a small tree"
- "a white column and a black fence... a potted plant"
- "a front porch... illuminated by a single light source"
- "a house with a porch and a tree... the number '01'"
- "a brick wall and a white column supporting a black roof"
Same scene, but each caption picks different features (a tree, a fence, a potted plant, a light source, a house number, a roof). These aren't near-duplicates by any text or embedding measure, so caption dedup can't collapse them. And the dhash gate didn't catch them either — the IR frames evidently differ enough (IR light, my movement, sensor grain) to clear the hamming-4 threshold.

Takeaway: for a live outdoor camera, per-frame content dedup (image-hash or caption) doesn't hold — consecutive frames genuinely differ. With Ring-MQTT holding motion "on" for the full 60–180s window (mine ran ~2.5 min) and the 3s loop, that's ~40+ genuinely-unique-looking frames per event, each firing a notification.

Suggestion: throttle at the notification/event level rather than per-frame — at most one "Camera Alert" per camera per motion event (or per N seconds), unless the scene changes materially, independent of caption/image similarity. Right now the "Camera Alert" path fires once per analysed snapshot with no event-level cap.

Also, still flagging: this path has no logging or audit trail. I checked the HA logs live during this storm — nothing from HGA at any level (an unrelated ESPHome device was flooding the log at the time). A little logging around whatever sends "Camera Alert" would let me capture the next one at code level. Happy to add it myself if you point me at the function.

(attached screenshot from my iPhone)
IMG_5369

goruck:
Thanks again for the continued testing — this is very helpful.

**Latest push** addresses the remaining issues you found:

---

### 1. Phantom notifications (recording-state without HA motion event)

The periodic recording-camera poll ran every 1.5 s regardless of whether an HA `binary_sensor.*_motion` event had fired. For Ring cameras, cloud-side detection puts the camera in `recording` state without triggering the HA motion sensor — so the poll fired freely.

**Fix:** The poll is now replaced with an event-driven approach. `_handle_camera_recording_state_change` starts a snapshot loop the moment a camera enters `recording` state (exactly what the motion loop does), and cancels it when the camera exits. If a `binary_sensor.*_motion` event also fires (Ring sends both), the motion event cancels the recording loop and takes ownership so the `motion OFF` event controls the lifecycle and final flush. The periodic poll remains as a safety net but skips any camera already covered by either loop.

---

### 2. Notification dispatch now visible in debug logs

Previously the debug log for notification dispatch only appeared in `notify_on_anomaly` mode (where the novelty decision is logged). In `always_notify` mode there was no log line. A `LOGGER.debug` is now emitted in `_send_notification` itself, so you will see it regardless of mode:

```
[<camera_name>] Dispatching notification via notify.<service>: "<caption>"
```

To enable debug logging for the integration, add to your `configuration.yaml`:

```yaml
logger:
  default: warning
  logs:
    custom_components.home_generative_agent: debug
```

---

### 3. VLM quality and semantic deduplication

The `notify_on_anomaly` mode suppresses repeated notifications by comparing new captions against recent ones via vector similarity. This only works reliably if your vision model produces **consistent captions** for the same scene.

Small models like `moondream` describe the same static scene differently on every call — so every frame appears "novel" and the dedup never fires. You'll see 42 notifications instead of 1.

**Recommendation:** Use at minimum a **7B-class VLM** such as `qwen2.5vl:7b` (recommended) or `llava:7b`. This is now documented in [camera-entities.md](../docs/camera-entities.md).

---

### 4. dHash uniqueness toggle (new UI option)

A new toggle — **"Enable perceptual hash (dHash) frame filter"** — is now available in Global Options. It is **off by default**.

**Important caveat before enabling:** dHash drops visually near-identical snapshots *before* they reach the VLM. This removes the temporal continuity the summary model uses to narrate motion across frames (e.g., "person approaches gate, then enters"). Only enable it if you are receiving excessive duplicate notifications from a static scene and you accept that inter-frame motion context may be lost.

If you're seeing many notifications, the root cause is more likely the VLM quality issue above — fix that first.

---

Please give this a try when you get a chance!

andymcmanus:
Thanks — this is a great set of changes, and #1 and #2 map exactly onto what I was seeing.

The event-driven recording-state loop is precisely the mechanism I couldn't pin down: my no-motion "phantom" bursts were Ring cameras entering recording via cloud-side detection with no HA binary_sensor.*_motion event, so the old periodic poll fired freely. Replacing it with a start-on-recording / cancel-on-exit loop (with the motion event taking ownership when it fires) is exactly right. I'll re-test and confirm.

The notification debug line in _send_notification is also very welcome — the total absence of logging on that path was why I could only ever send you screenshots. Being able to see [<camera>] Dispatching notification… in all modes means I can actually catch a burst in the act next time.

On the VLM: your diagnosis matches what my screenshot showed — moondream describes the same static scene differently every call, so dedup never fires. One constraint on my end: I'm running an integrated GPU (AMD 780M via ROCm), and it becomes unstable under heavier load — I've seen unspecified launch failure core dumps when pushing it. moondream was chosen partly for that reason. So rather than jump straight to a 7B VLM, I'm going to trial gemma3:4b first — enough of a step up from moondream that captions should be more stable for dedup, but far lighter than 7B on my hardware. If it isn't consistent or accurate enough I'll look at qwen2.5vl:7b, but I want to see whether a 4B gets me there within what this GPU can handle.

I'll report back on both the phantom/logging fixes and the VLM trial once I've tested on a clean feed. Thanks again — genuinely responsive fixes.

andymcmanus:
Re-tested the latest branch (f8fe464). First, the two structural fixes land well:

- The notification debug line in _send_notification is exactly what was needed — I can finally see this path (per-snapshot dispatch + the novelty decision). That instrumentation is what let me pin down what follows, so thank you.
- The event-driven recording loop behaves correctly: it starts when the camera's trigger goes active and cancels the instant it clears. The loop logic isn't the problem.

The problem is *which entity the resolver binds the camera to*, and it's worth explaining because it's a trap a very common Ring topology sets.

## The motion→camera resolver binds to the wrong sensor

On my setup, camera.front_door_snapshot is a Ring Battery Doorbell Plus. But there are THREE entities matching "front door motion":

- binary_sensor.front_door_motion   → a Ring **Alarm Motion Sensor** (an interior PIR, different device, mounted facing into the house)
- binary_sensor.front_door_motion_2 → the native `ring` integration's copy (being retired)
- binary_sensor.front_door_motion_3 → the doorbell's *own* motion (mqtt)

The resolver strips `_motion` and name-matches, so it binds camera.front_door_snapshot to binary_sensor.front_door_motion — which is the **interior alarm PIR**, not the doorbell. The doorbell's own motion sensor (front_door_motion_3) doesn't match the pattern and is ignored.

## What that causes

A single walk-out produced **62 snapshots → 62 notifications**. The debug log shows every frame flagged novel (more on that below), but the root cause of the *count* is the trigger: the alarm PIR was "on" from 16:09:47 to 16:13:05 — 3m18s — because an alarm PIR legitimately stays active while there's any movement in its zone (I walked out, came back, moved around). The snapshot loop faithfully sampled every 3s for that entire window. 198s ÷ 3s ≈ 62. The doorbell's own motion sensor stayed off the whole time, yet the Ring app *did* record the event — confirming the doorbell fired independently of the sensor the resolver chose.

This same mis-binding also explains the "phantom" bursts I reported earlier (camera alerts with nothing in the Ring activity log): interior movement past the hallway PIR triggers front-door camera analysis with no doorbell event behind it. It was never a cloud recording-state phantom — it's the resolver watching an interior alarm sensor.

## Deeper issue: motion binary_sensors are the wrong signal for battery doorbells

Even the *correct* sensor (front_door_motion_3) is unreliable here: battery-powered Ring cameras don't push motion/state promptly or reliably over MQTT. The signal that actually fires reliably, immediately, and only on a genuine Ring event is the **select.<camera>_event_select entity's `eventId` attribute**. My own automations trigger on eventId changes and have been rock-solid; the motion binary_sensors never were.

## Suggestions

1. **Resolve by device, not by name.** Prefer the motion binary_sensor that shares a device with the camera, rather than the first name-match. That alone would pick front_door_motion_3 over the unrelated alarm PIR.
2. **Let the trigger entity be user-overridable per camera.** Naming collisions between doorbells and alarm PIRs are common on Ring; an explicit override escape-hatch would cover the cases auto-resolution can't.
3. **Consider event_select `eventId` as the trigger signal for Ring-MQTT cameras** (at least battery ones), instead of / in addition to motion/state. It's the only signal I've found that's reliable on battery hardware and can't be tripped by a co-located alarm sensor.

Separately, the caption volume: with moondream, every one of those 62 frames was flagged notify=True reason=no_match — same static porch described differently each time, so dedup never fired. I'm going to trial gemma3:4b (moondream's too small to caption consistently, but a 7B is a stretch on my iGPU). But that's a duplicate-*suppression* issue and secondary to the trigger binding above — even perfect dedup would still run 62 VLM inferences per event, which the wrong-sensor binding is what actually drives.

goruck:
Thanks for the detailed debug log — the device/name collision explanation is exactly what was needed to root-cause this properly.

**Latest push (45e2798)** addresses both the resolver binding and the override escape-hatch:

---

### 1. Device-based motion→camera resolution

The resolver now uses a three-tier lookup:

1. **Explicit override map** — checked first (see below).
2. **Device registry** — finds the camera entity that shares the same HA device as the firing motion sensor. For your setup this means `binary_sensor.front_door_motion_3` (registered to the Ring doorbell device) correctly binds to `camera.front_door_snapshot`, while `binary_sensor.front_door_motion` (the alarm PIR, a different device entirely) is rejected automatically — no name matching involved.
3. **Name heuristics** — the existing direct-match, VMD-strip, and `_motion`-strip logic, reached only when tiers 1 and 2 produce nothing.

---

### 2. UI-configurable motion sensor override map

Global Options now has a **"Motion sensor → camera overrides"** multiline text field. One pair per line:

```
binary_sensor.front_door_motion_3: camera.front_door_snapshot
binary_sensor.backyard_sensor: camera.backyard
```

This is the explicit escape-hatch for cases the device registry can't resolve — name collisions, sensors from integrations that don't register a device, or any topology where you want to pin a specific sensor to a specific camera. Leave it blank and the device-based resolution handles it automatically.

---

The device-based approach should eliminate the phantom PIR→camera binding entirely without any manual configuration needed on your end. Let me know how it goes.

PR #511, opened 2026-07-27
fix(audit): correctly label triage-suppressed and policy-blocked find…

hruba202:
Title: Audit store: triage-suppressed and policy-blocked findings incorrectly recorded as not_suppressed

In sentinel/engine.py, findings suppressed after the initial should_suppress() check — via LLM triage (TRIAGE_SUPPRESS) or execution policy (ACTION_POLICY_BLOCKED, both simple and compound finding branches) — were still audited with suppression_reason_code=suppression_decision.reason_code (or a hardcoded "not_suppressed" for the compound case). Since suppression_decision.suppress was False at that point, reason_code was always SUPPRESSION_REASON_NOT_SUPPRESSED, mislabeling findings that never reached the user.

This caused AuditStore._is_evictable() to treat these as non-evictable, user-facing records, filling the hot buffer with functionally-suppressed noise and forcing eviction of genuinely user-facing history once audit_hot_max_records was reached.

Fix: added SUPPRESSION_REASON_TRIAGE_SUPPRESSED and SUPPRESSION_REASON_POLICY_BLOCKED constants in suppression.py, and updated the three affected _append_finding_audit() call sites in engine.py.


hruba202:
"Claude Code Review fails with an OIDC token error — this appears to be a known GitHub limitation for workflows triggered from forked-repo PRs (id-token: write isn't granted to fork PRs for security reasons), not an issue with the change itself

goruck:
Merged and released in [v3.21.1](https://github.com/goruck/home-generative-agent/releases/tag/v3.21.1) — thanks @hruba202, excellent diagnosis. Your analysis was exactly right on all three call sites, and the fix is in as-is.

Review added a few things on top (8641789):

- **Backfill for already-stored records** — records written before this fix still carried `not_suppressed` and would have stayed non-evictable forever (the buffer-clogging you found would persist on any install that already hit it, including yours). They're now relabeled automatically on load: a genuinely notified record can never carry `triage_decision == "suppress"` or `action_policy_path == "blocked"`, so the relabel is deterministic.
- **Response attribution** — with the new codes, a late dismiss/false-positive tap could land on a newer suppressed duplicate (same content-hash `anomaly_id`) and become invisible to KPIs and evictable. `async_update_response` now prefers the newest `not_suppressed` match.
- **Trigger-source KPI** — `notified_at` is stamped on every audit record, so the 24h trigger-source breakdown was counting suppressed findings too; it now filters on the reason code like the other notified KPIs.
- Tests for all three engine call sites, eviction of the new codes, the relabel, and response preference.

Two adjacent issues you didn't cause went to TODOS.md: notifier-internal drops (30-min duplicate gate / rate-limit batching) still audit as `not_suppressed` (P2, needs a delivery status from `async_notify`), and compound responses matching by any constituent id (P3).

Heads-up for your install: notified-KPI values and daily digest counts may show a one-time corrective drop after upgrading — that's the mislabeled records being counted correctly, not a regression. If you get a chance, I'd appreciate a sanity check after a few days of runtime that the audit hot buffer stops evicting genuine alert history on your setup.

The claude-review CI failure was the fork OIDC limitation you flagged — correct call, it's unrelated to the change.

Issue #518, opened 2026-07-29
[Rule] Unexpected Kitchen Motion While Away


hruba202:
### Rule name

unexpected_kitchen_motion_while_away

### Summary

Detects motion in the Kitchen area (binary_sensor.xiao_esp32_c5_espectre_motion) when no one is home.

### Motivation and safety impact

This proposal is currently marked unsupported in HGA and needs a deterministic template.
Candidate ID: motion_kitchen_while_away
Observed pattern: state_change

### Required snapshot evidence

- entities[31].state
- derived.anyone_home

### Deterministic detection logic

1) Evaluate only snapshot fields listed in required evidence.
2) Trigger when the candidate condition is true.
3) Return no findings when any required condition is missing.

### Evidence mapping

- evidence.path_1 <- entities[31].state
- evidence.path_2 <- derived.anyone_home

### Severity

low

### Confidence (0.0-1.0)

0.6

### Sensitive action?

- [ ] This rule is sensitive and must require confirmation for any action.

### Suggested actions (semantic only)

- close_entry

### Suppression/cooldown expectations

Use default per-type cooldown (30 min).

### Test cases

- Trigger: Detects motion in the Kitchen area (binary_sensor.xiao_esp32_c5_espectre_motion) when no one is home.
- Non-trigger: same snapshot context but with condition absent.

goruck:
Shipped in v3.23.0 (PR #521) — thanks @hruba202, this one turned out to be two features in disguise.

Your candidate's evidence paths were the index-based form (`entities[31].state`), which can never resolve to an entity ID — the index only means something against the snapshot the candidate was drafted from. So besides the `motion_detected_while_away` template itself, this release adds a prose-evidence fallback: the sensor named in the candidate summary (`binary_sensor.xiao_esp32_c5_espectre_motion`) is promoted as motion evidence, and approval verifies the ID actually exists in your HA instance before registering the rule (a hallucinated name is refused with `entities_unresolved` instead of silently registering a rule that can never fire).

Two deliberate deviations from the proposal, same spirit as #516:

- **`check_camera` instead of `close_entry`** — a motion finding has no deterministic door/window target, so nothing is actuated; the action stays advisory.
[...]

Issue #514, opened 2026-07-28
[Rule] Multiple Occupancy Sensors Unavailable


hruba202:
### Rule name

multiple_occupancy_sensors_unavailable

### Summary

Several occupancy sensors are simultaneously unavailable, potentially indicating a system outage.

### Motivation and safety impact

This proposal is currently marked unsupported in HGA and needs a deterministic template.
Candidate ID: multiple_occupancy_sensors_unavailable
Observed pattern: state_unavailable

### Required snapshot evidence

- entities[entity_ids contains "binary_sensor.0x00124b0010b0a987_occupancy"].state
- entities[entity_ids contains "binary_sensor.0xfc012cfffef62ae0_occupancy"].state
- entities[entity_ids contains "binary_sensor.smart_presence_sensor_obsazenost"].state

### Deterministic detection logic

1) Evaluate only snapshot fields listed in required evidence.
2) Trigger when the candidate condition is true.
3) Return no findings when any required condition is missing.

### Evidence mapping

- evidence.path_1 <- entities[entity_ids contains "binary_sensor.0x00124b0010b0a987_occupancy"].state
- evidence.path_2 <- entities[entity_ids contains "binary_sensor.0xfc012cfffef62ae0_occupancy"].state
- evidence.path_3 <- entities[entity_ids contains "binary_sensor.smart_presence_sensor_obsazenost"].state

### Severity

low

### Confidence (0.0-1.0)

0.3

### Sensitive action?

- [ ] This rule is sensitive and must require confirmation for any action.

### Suggested actions (semantic only)

- close_entry

### Suppression/cooldown expectations

Use default per-type cooldown (30 min).

### Test cases

- Trigger: Several occupancy sensors are simultaneously unavailable, potentially indicating a system outage.
- Non-trigger: same snapshot context but with condition absent.

goruck:
Shipped in [v3.21.2](https://github.com/goruck/home-generative-agent/releases/tag/v3.21.2) (PR #515) — thanks @hruba202 for the well-formed proposal, the structured template-request format made this straightforward to pin down.

The root cause was normalization-only: the unavailable-sensors templates collected `sensor.*` evidence IDs but your candidate cites `binary_sensor.*` occupancy/presence entities. Both domains are accepted now, and the review pass hardened the surrounding routing while we were in there:

[...]

Issue #504, opened 2026-07-24
[Rule] Window Open Detected at Night


hruba202:
### Rule name

window_open_detected_at_night

### Summary

The bedroom window was opened during night hours.

### Motivation and safety impact

This proposal is currently marked unsupported in HGA and needs a deterministic template.
Candidate ID: window_open_at_night
Observed pattern: state_change

### Required snapshot evidence

- entities[entity_ids contains 'binary_sensor.bedroom_loznice_okno'].state
- derived.is_night

### Deterministic detection logic

1) Evaluate only snapshot fields listed in required evidence.
2) Trigger when the candidate condition is true.
3) Return no findings when any required condition is missing.

### Evidence mapping

- evidence.path_1 <- entities[entity_ids contains 'binary_sensor.bedroom_loznice_okno'].state
- evidence.path_2 <- derived.is_night

### Severity

low

### Confidence (0.0-1.0)

0.5

### Sensitive action?

- [ ] This rule is sensitive and must require confirmation for any action.

### Suggested actions (semantic only)

- close_entry

### Suppression/cooldown expectations

Use default per-type cooldown (30 min).

### Test cases

- Trigger: The bedroom window was opened during night hours.
- Non-trigger: same snapshot context but with condition absent.

goruck:
Shipped in [v3.21.0](https://github.com/goruck/home-generative-agent/releases/tag/v3.21.0) (PR #512) — thanks @hruba202 for the well-specified template request.

Your candidate was actually hitting three separate problems, all fixed:

1. **Quoted entity IDs** — the evidence path `entities[entity_ids contains 'binary_sensor.bedroom_loznice_okno'].state` kept the quotes attached during extraction, so the entity was never recognized. Quote-wrapped IDs now resolve everywhere (normalization, camera matching, duplicate detection).
2. **Czech entity naming** — entry detection only knew the English tokens "door"/"window"/"entry", so `okno` couldn't match. When the candidate's English description names a door or window, locale-named binary sensors and covers are now promoted to entry sensors.
3. **No matching template** — "opened during night hours" was also tripping the duration heuristic, and there was no template for night + unknown occupancy. There's now a dedicated presence-agnostic **`open_entry_at_night`** template: it fires when the entry is open and `derived.is_night` is true, whether or not anyone is home — exactly your proposal's evidence spec.

[...]

PR #523, opened 2026-07-30
feat: optional response language for Sentinel discovery and triage text

hruba202:
Discovery candidate title/summary and triage summary are always LLM-generated in English, regardless of the HA UI language — cs.json only covers static integration strings, not this free text.

Adds sentinel_response_language (Sentinel subentry, optional, default empty = English) that appends a language instruction to the discovery and triage prompts, mirroring vlm_response_language's convention: only free-text fields (title, summary) are translated; every machine-matched field (candidate_id, pattern, evidence_paths, decision, reason_code, schema_version, etc.) is explicitly instructed to stay in its defined form and is never touched by this option.

- const.py: new CONF_SENTINEL_RESPONSE_LANGUAGE / RECOMMENDED default
- explain/discovery_prompts.py: LANGUAGE_INSTRUCTION_TEMPLATE, appended to the user prompt only when the option is set
- sentinel/discovery_engine.py: reads the option from runtime options
- sentinel/triage.py: SentinelTriageService gains a response_language param and its own language instruction appended to the system prompt
- __init__.py: wires the option into SentinelTriageService
- flows/sentinel_subentry_flow.py: new optional text field in the Sentinel subentry form
- core/subentry_resolver.py: added the key to the _apply_sentinel_options allowlist -- without this the option is stored but never reaches the engine/triage service (same class of bug flagged as critical in #480)
- strings.json / translations/en.json / translations/cs.json: field label and description
- docs/sentinel.md, docs/constants.md: documented the new option

Tested end-to-end against a live v3.21.3 install: discovery candidate title/summary and triage summary render in the configured language while candidate_id/pattern/decision/reason_code stay untranslated.


goruck (changes requested):
Thanks @hruba202 — this is another carefully built PR. The plumbing is genuinely flawless: you pre-empted the #480 allowlist bug class yourself, the flow persistence works (including clearing the field back to English), and the machine-field carve-outs in both prompt templates are exactly the right instinct. I also appreciate that you field-tested it end-to-end before submitting, and that you filed #524 for the card-side gap you spotted while building this.

Unfortunately I can't merge it in its current shape — and #524 turns out to be the visible tip of exactly why.

**The discovery leg: candidate `title`/`summary` aren't display text — they're a machine surface.** What you found in the card (English regex badge detection, #524) also lives server-side, where the consequences aren't cosmetic: the normalization and dedup layers parse English phrases out of that prose, and the dependency is deliberate. There's an explicit comment in `proposal_templates.py` (from your own #504 fix, ironically): *"Entity IDs follow the user's locale... The candidate text is always English — fall back to it."* Your Czech entity IDs work **because** the prose is English. Concretely, with `sentinel_response_language: Czech` we reproduced during review:

1. **Silent home/away inversion.** An "open while someone is home" candidate with Czech prose and bare occupancy evidence normalizes to `open_entry_while_away` — the occupancy-scoping regexes only match English, so the fallback picks the away template. The approved rule watches the *opposite* condition. The lock case inverts the other way. This is the worst failure mode: not a rejection, a wrong rule.
2. **Unapprovable candidates.** Your own #504 window case (`binary_sensor.bedroom_loznice_okno`) normalizes to `open_entry_at_night` with English prose but becomes `unsupported_pattern` with Czech prose — approval re-runs normalization on the stored text, so the approve button returns "unsupported".
3. **Dedup breaks.** `discovery_semantic.py` derives predicate and home/away key components from the same English terms (the machinery hardened through #514–#518). Czech candidates never dedup against their own activated rules → repeated proposals; switching languages also changes the title+summary hash fallback.
4. **The evidence-mismatch guard goes silent.** `_candidate_text_entity_mismatches` token-matches prose against entity object-IDs with an ASCII tokenizer — translated prose produces no matches, so a candidate could describe one device while its evidence targets another, undetected.

Your live test passed because the happy path — Czech titles rendering, machine fields intact — really does work. The breakage lives downstream in normalization, dedup, and approval, which is exactly what makes this class of issue dangerous.

**On #524 itself:** your structured-evidence direction is the right one, and it's bigger than the card — making away/home/night context primarily `evidence_paths`-driven (with the English regexes demoted to legacy fallback) is the *precondition* for ever translating discovery prose at generation time, server-side included. Two notes: current `main` (v3.22.0+) already gives `isAway` one structured path — `evidencePaths.includes("not derived.anyone_home")` (added in PR #520; your quoted snippet predates it) — though it only fires when the model emits that exact path, so your point stands. And your alternative (gating the text fallback on `sentinel_response_language` being unset) doesn't hold server-side: candidates persist across option changes, so a stored Czech candidate stays Czech after the option is cleared. Let's use #524 as the design home for the structured-evidence work.

**The triage leg: please drop it, and this one's on us.** Two things your PR uncovered: `TriageDecision.summary` is parsed and then never consumed anywhere (its only appearance is a debug log — no notification, audit record, or card renders it), and `sentinel_triage_enabled` itself has no config-flow field and isn't in the resolver allowlist, so the triage service can't actually be enabled from the UI at all. Our `docs/sentinel.md` presents it as a normal option, which I suspect is what pointed you at it — sorry about that. Both are now tracked as TODOs on our side (expose-or-remove decision for triage, plus the dead summary field, plus the docs fix).

**The good news: there's a better target, and your plumbing mostly carries over.** The sentinel text a Czech user actually reads — the finding explanation in mobile and persistent notifications — comes from the *explainer* (`explain/prompts.py` → `LLMExplainer`, the "Enable LLM explanations for sentinel findings" switch). That output is display-only: nothing machine-parses it, so it's the safe, high-value place for `sentinel_response_language`. If you rework the PR:

- Keep your const/flow/resolver/translations plumbing essentially as-is — just point the language instruction at the explainer prompt instead of discovery/triage.
- One hardening requirement: `_redact_if_sensitive` strips recognized-person names from explanations on sensitive findings by exact match. Czech declension inflects names ("Petra" → "Petru"), so the language instruction must tell the model to keep person names in their base (nominative) form exactly as given — otherwise a name could leak through redaction. A test with an inflected name would lock that in.
- Please also drop the stray whitespace-only line that slipped into `translations/en.json` (line 335 — `git diff --check` flags it, and en.json must mirror strings.json byte-for-byte; it's test-enforced), and add tests along the lines of `test_vlm_prompt_overrides.py`.

So the shape I'd propose: this PR becomes the explainer-language option (small rework, I'll review it fast), #524 becomes the structured-evidence design issue, and once that lands, generation-time translation of discovery prose becomes a safe follow-up instead of a foot-gun.

Sorry to push back on something you already have running live. Your PRs keep finding the sharp edges of this integration — this one surfaced a dormant feature and a dead field we didn't know we had, and #524 is a genuinely useful map of the card-side gap.

hruba202:
Thanks for the detailed review — reworked per your feedback in `5d9ba0d` (rebased onto current `main`, no conflicts, CI green). Summary of changes:

**Discovery leg — reverted.** `discovery_prompts.py` and `discovery_engine.py` are back to upstream; `LANGUAGE_INSTRUCTION_TEMPLATE` and its call site are gone. Candidate `title`/`summary` stay English-only, so normalization, dedup, and the evidence-mismatch guard are unaffected.

**Triage leg — dropped entirely**, as suggested. `triage.py` and the `SentinelTriageService` constructor are back to upstream (no `response_language` param, no appended instruction).

**Retargeted to the explainer**, per your proposal:
- `explain/prompts.py`: new `LANGUAGE_INSTRUCTION_TEMPLATE`, appended to `SYSTEM_PROMPT` only when `sentinel_response_language` is set — mirrors `agent/tools.analyze_image`'s VLM language override.
- `explain/llm_explain.py`: `LLMExplainer.__init__` takes `response_language: str = ""`; `async_explain` appends the instruction to the system prompt when set.
- `__init__.py`: `response_language=` wiring moved from `SentinelTriageService(...)` to `LLMExplainer(...)`.
- `sentinel_subentry_flow.py`: the `CONF_SENTINEL_RESPONSE_LANGUAGE` field (and its `_default_payload` entry) moved to sit next to `CONF_EXPLAIN_ENABLED` instead of the discovery options — it's the explainer's setting now, not discovery's.

**Nominative-name hardening (your point on `_redact_if_sensitive`).** The language instruction explicitly requires the model keep recognized person names in their base/nominative form, uninflected, "even where {language} grammar would normally require this." Locked in with two paired tests in `test_llm_explain.py`:
- `test_redaction_succeeds_when_model_keeps_nominative_name` — a compliant explanation (name kept nominative) redacts correctly even with surrounding Czech declension.
- `test_redaction_misses_inflected_name_documents_the_gap` — documents that an inflected name (`Petra` → `Petru`) would slip past `_redact_if_sensitive`'s exact match, which is exactly why the instruction is a hard requirement rather than a nice-to-have.

Also added `test_no_language_override_leaves_system_prompt_unchanged`, `test_language_override_appends_to_system_prompt`, and `test_language_override_instructs_nominative_person_names` — same shape as `test_vlm_prompt_overrides.py`.

**`en.json` line 335.** Stray blank line removed; `strings.json` and `translations/en.json` are verified byte-for-byte identical and `git diff --check` is clean.

**Docs.** `docs/constants.md` and `docs/sentinel.md` updated to describe the explainer scope, with a pointer to #524 for the discovery-side structured-evidence follow-up.

Ruff check + format both pass clean on every touched file, and the branch now rebases onto current `main` (through #526/#528) without conflicts. Let me know if anything else needs adjusting.

goruck (approved):
Approving — the rework lands exactly where the review discussion pointed, and it's cleanly done.

**Verified locally against `origin/main`:**
- `discovery_prompts.py`, `discovery_engine.py`, and `triage.py` are byte-identical to main — the discovery/triage legs are fully reverted, so normalization, dedup, and the evidence-mismatch guard keep operating on English prose, and triage `decision`/`reason_code` stay machine-matched. The #524 follow-up remains the home for making discovery prose translatable safely.
- The option now flows config flow → subentry resolver allowlist → `LLMExplainer` only, and the explainer is the right target: its output goes to notifications and raw audit storage, and nothing downstream parses it (checked all consumers).
- Full suite, ruff, and pyright green; your five new tests are good, and the field placement next to `explain_enabled` is the right call.

**One hardening commit added on top (0662803)** — I hope you don't mind the direct push (maintainer edits were enabled). Cross-model adversarial review flagged one gap in the redaction story: the nominative-name instruction can't be the *only* thing standing between an inflected name and a leaked notification, because prompt compliance is never guaranteed — your own `test_redaction_misses_inflected_name_documents_the_gap` demonstrates the failure honestly. The project invariant is that the LLM never gates safety behavior, and redaction is safety behavior.

So the commit makes the boundary deterministic: when `sentinel_response_language` is set and a finding is sensitive, recognized names are redacted from the **evidence structure before the prompt is rendered** (`_redact_person_names`) — a name the model never sees cannot be emitted in any inflection. Your instruction stays in the prompt as defense in depth, and it still keeps names natural-looking in non-sensitive findings.

Three details in the implementation that took adversarial rounds to get right, recorded here for posterity:
1. **Longest-first name matching** — with `recognized_people=["Alex", "Alexander"]`, a naive alternation turns "Alexander" into "a recognised personander"; the residue still identifies the person.
2. **Structural (pre-render) redaction, not post-render string redaction** — Python's repr escapes quotes, so a name like `D'Angelo` wouldn't match its rendered form; and a post-render pass on a short name like "Max" would corrupt the template's own "Max 2 short sentences." rule.
3. **The walk covers dict keys, tuples, sets, and frozensets**, not just values and lists, and never mutates the original evidence (its key order feeds anomaly-id derivation).

English behavior is unchanged — with no response language set, the prompt is not pre-redacted and the existing post-hoc exact-match redaction applies as before.

Also fixed in the same commit: the `subentry_resolver.py` comment still described the pre-rework discovery/triage scope.

1929 tests pass (your 5 + 7 new redaction regressions). Merging once CI confirms. Thanks for the disciplined rework — reverting the two legs completely rather than trying to salvage them made this easy to verify, and the paired redaction tests documenting the gap honestly are exactly the right instinct. Ty děkujeme!

goruck:
Shipped in [v3.24.0](https://github.com/goruck/home-generative-agent/releases/tag/v3.24.0) 🎉

When you get a chance to run it against your live install, two things worth confirming in the field:
1. Finding explanations arrive in Czech in both mobile and persistent notifications, and discovery/triage text stays English as designed.
2. For a **sensitive** finding (recognized person on camera), the notification shows *a recognised person* — never a name, in any declension. With the pre-redaction the model never sees the name at all, so the explanation should read naturally around the generic phrase.

No rush — thanks again for the contribution.

hruba202:
Ran it against my live install, here's what I found:

1. ⚠️ Not able to cleanly confirm yet — the "no person in image" notification I initially checked turned out to be coming from one of my own automations, not the actual Sentinel finding pipeline, so it's not a valid test. Will retest against a genuine Sentinel finding and report back.

2. Redaction is not happening for recognized persons — but raising this as a design question rather than a pure bug. Got a sensitive Sentinel finding notification with:
   - Summary: "V obrázku je osoba." (correct, Czech)
   - Recognized: **h202** (actual name)

   Per the PR description this is unexpected — the explanation is supposed to read naturally around a generic "recognised person" phrase. In practice though, having the real name in the "Recognized" field is actually useful for me: I have automations keyed off it, e.g. `recognized person: h202 → turn on living room music`, `recognized person: person 2 → turn on kid's room light`. If redaction becomes strict, that pattern breaks.

   So: is the name only meant to be scrubbed from the LLM-facing explanation text (to avoid it leaking into the generated prose), while the structured "Recognized" field stays intact for automation use? If so, what I'm seeing might actually be correct behavior, not a bug. Would be good to clarify the intended scope.

3. Found a real bug, tracked it down: **low-severity "Home Update" findings (baseline_deviation, time_of_day_anomaly, entity_staleness, appliance_power_duration, alarm_disarmed_open_entry) always render in English, regardless of `response_language`.**

   Root cause is in `sentinel/notifier.py`, `_mobile_message()`:
   ```python
   formatter = _TEMPLATE_MOBILE_FORMATTERS.get(
       str(finding.evidence.get("template_id") or "")
   )
   if formatter:
       return formatter(finding)
   if explanation:
       
   ```
   For these five template_ids, a deterministic hardcoded-English formatter (e.g. `_baseline_deviation_mobile_message`) runs *instead of* the translated LLM `explanation`, so `response_language` never gets a chance to apply. Example from my install:
   > "hfoun: 1.0W vs usual X.XW (below normal). Check appliance."

   `_persistent_message()` has the same issue in its own fallback branch when explanation is empty/None.

   This looks like intentional deterministic copy for reliability/cost reasons, so probably not a quick fix — likely needs either translated string templates for these formatters, or routing them through the LLM explainer when `response_language` is set. Happy to open a separate issue/PR for this if useful, since it's outside what #523 set out to do.

Screenshots available if useful.

goruck:
Thanks — that's a genuinely useful field report, and point 3 is a real bug. Taking each in turn.

**1. Retest.** No problem, and good catch on the source. Ping me when you have a genuine Sentinel finding to check against.

**2. Redaction scope — your read is correct, this is intended behavior.** Redaction is deliberately prose-only. It exists to stop a recognized person's name from being *narrated* into LLM-generated text, not to hide the name from your automations.

Two layers, both scoped to the explanation:

- `explain/llm_explain.py` `_redact_person_names` strips names from the evidence *before* the model sees them (so no inflection can leak one back)
- `sentinel/notifier.py` `_redact_if_sensitive` scrubs the generated text as a second pass

Neither touches the snapshot evidence, the `recognized_people` sensor state, or its attributes. Your `recognized person: h202 → turn on living room music` automations are safe, and will stay that way — keying automations off the structured field is exactly what it's there for.

Worth noting where your "Recognized: h202" line actually came from: the Sentinel notifier never emits a `Recognized:` field. That's `sensor.<camera>_recognized_people` (`core/recognized_sensor.py`), so what you saw was almost certainly rendered by one of your own automations — consistent with what you found in point 1. A genuine sensitive Sentinel notification has only title, subtitle, and message, and the name would only ever be at risk in that message.

I've made the scope explicit in `docs/sentinel.md` so the next person doesn't have to ask.

**3. Confirmed, and it's slightly worse than you found.** Your root-cause analysis is exactly right. Six finding types, not five — `alarm_disarmed_during_external_threat` has the same problem, dispatched on `finding.type` rather than `template_id` alongside `appliance_power_duration`.

The part you didn't spot: `_persistent_message()` already prefers the explanation. So for one and the same finding, the **mobile push is English while the persistent notification is Czech**. It isn't only a missing translation, it's the two surfaces disagreeing.

You're right that the deterministic copy is intentional — those formatters quote exact figures, units, and entity names that an LLM paraphrase blurs. But that reasoning only holds when the notification is meant to be English. Once you've set a response language, an English deterministic string is the one part of the notification ignoring your setting.

Fixed in #531: when a response language is set, the mobile push prefers the translated explanation, matching the persistent notification. With no language set, the deterministic copy still wins, so English installs are unchanged. When no usable translation exists (explainer off, or a translation over the 220-char mobile cap), it falls back to the deterministic string rather than the generic one — accurate English beats vague English.

One thing #531 does *not* fix: the notification **title** (`Security Alert` / `Home Alert` / `Home Update`) and the **subtitle** line are fixed strings rather than model output, so they stay English. Translating those needs real translated string templates, which is a larger piece of work than this fix. It's documented rather than half-done.

No need to open a separate issue — but if you want to take the title/subtitle translation, that one's genuinely open and I'd review it.

Screenshots would still be useful for the point-1 retest if it turns up anything odd.

goruck:
Correction to my last comment on point 3 — I said all six types would defer to the translation. Cross-model review pushed back on two of them and I think it's right, so #531 shipped narrower than I described.

**The four informational types translate as promised:** `baseline_deviation`, `time_of_day_anomaly`, `entity_staleness`, `appliance_power_duration`. That covers the case you actually hit.

**The two alarm types stay English:** `alarm_disarmed_during_external_threat` and `alarm_disarmed_open_entry` — so the one you listed as `alarm_disarmed_open_entry` will still arrive in English. Here's the reasoning.

That deterministic copy reads:

> Backyard detected unrecognized outdoor activity 2 min ago. The alarm has been disarmed since 10:15 PM. Arm the alarm or view the camera.

The LLM paraphrase in our own regression test reads:

> The home alarm was recently disarmed while someone is still inside.

No camera, no time, no call to action — and it asserts someone is inside, which the evidence never established. There's a test that has pinned that exact string as must-not-appear since before your report; my first cut quietly unpinned it whenever a language was set.

For an informational power reading, a paraphrase that blurs a number is a cosmetic loss. For a security event, dropping *which* entry and *what to do* is not, and I'd rather you get precise English than readable prose that lost the actionable part. Proper translation for these needs real string templates rather than model output — that's a genuine follow-up, not a no.

Two other things that fell out of the same review, both worth knowing for your install:

**Over-long translations were degrading to worse English, not better.** The explainer caps output at 220 characters and substituted an English compact fallback (`"Open entry at night: Front Door. Urgent: check and secure it now."`). That fits the mobile cap, so it outranked the precise deterministic copy. Net effect: a Czech user whose translation ran long got *vaguer* English than before. Fixed — with a language set the explainer now reports the failure instead, and the precise copy wins.

**Batching.** If more than three non-high-severity notifications fire within 60 seconds they collapse into a single English `"N home updates: …"` summary and the per-finding text is never delivered. Same for the daily digest, and for notification titles and subtitles. All fixed strings, all still English, all now documented in `docs/sentinel.md` so it's not a surprise.

Two things worth checking when you next test, beyond the original retest:

1. A power/baseline finding arrives in Czech on **both** the phone and the persistent notification (they used to disagree).
2. An `alarm_disarmed_open_entry` finding arrives in **English** — expected now, not a regression.

Thanks for the precise root-cause on this one. Pointing straight at `_mobile_message` is what made it a same-day fix.

goruck:
Shipped in [v3.25.1](https://github.com/goruck/home-generative-agent/releases/tag/v3.25.1) 🎉 — your `_mobile_message` root-cause plus two more problems that surfaced while fixing it.

Three things worth confirming in the field when you upgrade:

1. **An informational finding arrives in Czech on both surfaces.** A `baseline_deviation` or `entity_staleness` finding should now read the same language on the phone push and in the persistent notification. Those two used to disagree, which is the part of the bug that wasn't in your report.
2. **An `alarm_disarmed_open_entry` finding arrives in English.** Expected, not a regression — see my previous comment for why security copy stays deterministic.
3. **Over-long explanations no longer degrade.** If a Czech explanation ever ran past 220 characters you were previously getting a *vaguer* English string than before the feature existed, because the explainer's compact fallback outranked the precise deterministic copy. That's fixed; you should now get the precise English copy in that case rather than the generic one.

One behaviour that is unchanged but worth knowing about, since it can look like a bug: if more than three non-high-severity findings fire within 60 seconds they collapse into a single English `"N home updates: …"` push and the per-finding text isn't delivered at all. Same for the daily digest and for notification titles and subtitles. Those are fixed strings built from counts and type labels rather than model output, so there's nothing to translate without real string templates. All documented in `docs/sentinel.md` now.

Point 1 from your report (the retest against a genuine Sentinel finding) is still open whenever you get to it — no rush.

Thanks again. Pointing straight at the formatter dispatch is what made this a same-day turnaround, and the two follow-on problems would not have been found without your report.

Issue #428, opened 2026-05-31
Chat not available with gemini


serhatozkara:
## System Information

version | core-2026.5.4
-- | --
installation_type | Home Assistant OS
dev | false
hassio | true
docker | true
container_arch | amd64
user | root
virtualenv | false
python_version | 3.14.2
os_name | Linux
os_version | 6.12.85-haos
arch | x86_64
timezone | Europe/Athens
config_dir | /config

<details><summary>Home Assistant Community Store</summary>

GitHub API | ok
-- | --
GitHub Content | ok
GitHub Web | ok
HACS Data | ok
GitHub API Calls Remaining | 5000
Installed Version | 2.0.5
Stage | running
Available Repositories | 3123
Downloaded Repositories | 15

</details>

<details><summary>Home Assistant Cloud</summary>

logged_in | false
-- | --
can_reach_cert_server | ok
can_reach_cloud_auth | ok
can_reach_cloud | ok

</details>

<details><summary>Home Assistant Supervisor</summary>

host_os | Home Assistant OS 17.3
-- | --
update_channel | stable
supervisor_version | supervisor-2026.05.1
agent_version | 1.8.1
docker_version | 29.3.1
disk_total | 30.8 GB
disk_used | 19.3 GB
nameservers | 192.168.155.1
healthy | true
supported | true
host_connectivity | true
supervisor_connectivity | true
ntp_synchronized | true
virtualization | kvm
board | ova
supervisor_api | ok
version_api | ok
installed_addons | Get HACS (1.3.1), Matter Server (8.5.0), Mosquitto broker (7.1.0), File editor (6.0.0), Omada Controller Stable (5.15.24.19), Let's Encrypt (6.3.2), Zigbee2MQTT (2.10.1-1), Nginx Proxy Manager (2.1.0), Duck DNS (2.0.0), ESPHome Device Builder (2026.5.1), Guacamole Client (1.6.0-bullseye-3), OpenThread Border Router (3.0.0), Matterbridge (2026.5.4), PostgreSQL with pgvector (0.1.0)

</details>

<details><summary>Dashboards</summary>

dashboards | 4
-- | --
resources | 3
views | 2
mode | storage

<details><summary>Recorder</summary>

oldest_recorder_run | April 28, 2026 at 7:49 PM
-- | --
current_recorder_run | May 31, 2026 at 1:14 PM
estimated_db_size | 429.67 MiB
database_engine | sqlite
database_version | 3.49.2

</details>

[...]

PR #527, opened 2026-08-01
fix: sanitize invalid 'required' in anyOf branches for Gemini tool schemas

hruba202:
## Summary

Fixes a Gemini 400 error on tool declarations whose JSON schema contains
an `anyOf` branch carrying an invalid `required`:

```
GenerateContentRequest.tools[0].function_declarations[1].parameters.any_of[0].required: only allowed for OBJECT type
GenerateContentRequest.tools[0].function_declarations[1].parameters.any_of[0].required[0]: property is not defined
... (repeated for any_of[1], any_of[2])
```

**Root cause:** `voluptuous_openapi` (`vol.Any(...)`) and pydantic
`Optional`/`Union` schemas can emit `anyOf` branches where a `required`
array is present on a non-`object` branch, or references a property that
isn't defined in that specific branch's `properties`. Gemini's protobuf
validation rejects both shapes. `_ensure_array_items()` (added in #430)
already normalizes the sibling `items`/`anyOf` issue but doesn't touch
`required`.

**Fix:** `_sanitize_any_of_required()`, called alongside
`_ensure_array_items()` in `_format_and_dedupe_tools()`:
- Strips `required` from any `anyOf` variant that isn't `type: object`.
- For `object` variants, keeps only the `required` entries that are
  actually defined in that variant's `properties`; drops the key
  entirely if none remain.
- Recurses into `properties` and `items`, same traversal pattern as
  `_ensure_array_items()`.
- No-op for OpenAI/Anthropic (only triggers when `anyOf` + `required`
  co-occur in that specific invalid shape).

## Test Coverage

9 new unit tests in `test_graph_helpers.py`:
- `required` stripped from non-object anyOf variant
- Undefined property stripped from object variant's `required`
- `required` key removed entirely when empty after filtering
- Valid object variant with well-formed `required` left unchanged
- Recursion into nested `properties`
- Recursion into `items`
- No-op when no `anyOf` present
- End-to-end `_format_and_dedupe_tools` reproducing the exact reported
  3-branch schema shape from the traceback

## Test plan
- [x] Unit tests pass (verified schema-transform logic in isolation)
- [x] Verified against a live HA instance with the Gemini provider —
      the tool call that previously 400'd now succeeds

## Related
Same failure class as #428/#430 (Gemini rejecting HGA's normalized tool
schemas), different manifestation — `items` vs `required`.

### DIFF
--- a/custom_components/home_generative_agent/agent/graph.py
+++ b/custom_components/home_generative_agent/agent/graph.py
@@ -1236,8 +1236,7 @@ def _ensure_array_items(schema: dict[str, Any]) -> dict[str, Any]:
 
 def _sanitize_any_of_required(schema: dict[str, Any]) -> dict[str, Any]:
     """
-    Strip 'required' from anyOf branches that aren't type:object, or whose
-    required properties aren't defined in that branch's 'properties'.
+    Strip invalid 'required' entries from anyOf branches.
 
     Gemini rejects 'required' on non-OBJECT anyOf variants, and rejects
     required entries that reference an undefined property. voluptuous_openapi
@@ -1246,22 +1245,24 @@ def _sanitize_any_of_required(schema: dict[str, Any]) -> dict[str, Any]:
     """
     if "anyOf" in schema:
         new_variants = []
-        for variant in schema["anyOf"]:
-            if isinstance(variant, dict):
-                variant = _sanitize_any_of_required(variant)
-                if "required" in variant:
-                    props = variant.get("properties", {})
-                    if variant.get("type") != "object":
-                        variant = {
-                            k: v for k, v in variant.items() if k != "required"
+        for raw_variant in schema["anyOf"]:
+            variant = raw_variant
+            if isinstance(raw_variant, dict):
+                sanitized = _sanitize_any_of_required(raw_variant)
+                if "required" in sanitized:
+                    props = sanitized.get("properties", {})
+                    if sanitized.get("type") != "object":
+                        sanitized = {
+                            k: v for k, v in sanitized.items() if k != "required"
                         }
                     else:
-                        req = [r for r in variant["required"] if r in props]
-                        variant = (
-                            {**variant, "required": req}
+                        req = [r for r in sanitized["required"] if r in props]
+                        sanitized = (
+                            {**sanitized, "required": req}
                             if req
-                            else {k: v for k, v in variant.items() if k != "required"}
+                            else {k: v for k, v in sanitized.items() if k != "required"}
                         )
+                variant = sanitized
             new_variants.append(variant)
         schema = {**schema, "anyOf": new_variants}
     if "properties" in schema:
--- a/custom_components/home_generative_agent/agent/graph.py
+++ b/custom_components/home_generative_agent/agent/graph.py
@@ -1234,6 +1234,48 @@ def _ensure_array_items(schema: dict[str, Any]) -> dict[str, Any]:
     return schema
 
 
+def _sanitize_any_of_required(schema: dict[str, Any]) -> dict[str, Any]:
+    """
+    Strip 'required' from anyOf branches that aren't type:object, or whose
+    required properties aren't defined in that branch's 'properties'.
+
+    Gemini rejects 'required' on non-OBJECT anyOf variants, and rejects
+    required entries that reference an undefined property. voluptuous_openapi
+    (vol.Any(...)) and pydantic Optional/Union schemas can emit both, e.g.
+    ``{"anyOf": [{"required": ["x"]}, {"type": "array", ...}]}``.
+    """
+    if "anyOf" in schema:
+        new_variants = []
+        for variant in schema["anyOf"]:
+            if isinstance(variant, dict):
+                variant = _sanitize_any_of_required(variant)
+                if "required" in variant:
+                    props = variant.get("properties", {})
+                    if variant.get("type") != "object":
+                        variant = {
+                            k: v for k, v in variant.items() if k != "required"
+                        }
+                    else:
+                        req = [r for r in variant["required"] if r in props]
+                        variant = (
+                            {**variant, "required": req}
+                            if req
+                            else {k: v for k, v in variant.items() if k != "required"}
+                        )
+            new_variants.append(variant)
+        schema = {**schema, "anyOf": new_variants}
+    if "properties" in schema:
+        schema = {
+            **schema,
+            "properties": {
+                k: _sanitize_any_of_required(v) for k, v in schema["properties"].items()
+            },
+        }
+    if isinstance(schema.get("items"), dict):
+        schema = {**schema, "items": _sanitize_any_of_required(schema["items"])}
+    return schema
+
+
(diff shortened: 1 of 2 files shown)


goruck (changes requested):
Thanks for this one — the root-cause analysis is exactly right (same failure class as #428/#430, `required` instead of `items` this time), the repro test that rebuilds the 3-branch traceback shape is great, and mirroring the `_ensure_array_items()` traversal keeps the two normalizers easy to reason about side by side. The live-Gemini verification in the test plan is appreciated as always.

I have to request changes though, because of one structural issue: **the sanitizer runs for every provider, and the exact shape it strips is a real, load-bearing constraint for the non-Gemini ones.**

## P1 — unconditional stripping deletes HA's at-least-one-target constraint for OpenAI/Anthropic/Ollama

voluptuous_openapi emits the standard HA "must supply at least one target" pattern — `vol.Required(vol.Any("name", "area", "floor"))` on intent tools like `HassTurnOn` — as parent-level `properties` plus bare-`required` branches (`voluptuous_openapi/__init__.py`, the `itertools.product(*any_of_constraint_groups)` block):

```python
import voluptuous as vol
from voluptuous_openapi import convert

schema = vol.Schema({
    vol.Required(vol.Any("name", "area", "floor")): str,
    vol.Optional("domain"): str,
})
convert(schema)
```

```json
{
  "type": "object",
  "properties": {"name": {...}, "area": {...}, "floor": {...}, "domain": {...}},
  "anyOf": [
    {"required": ["name"]},
    {"required": ["area"]},
    {"required": ["floor"]}
  ]
}
```

Those bare-`required` branches are *valid* JSON Schema — implicit-object constraints against the parent's properties. Only Gemini's protobuf validation rejects them (it's also, I'm fairly sure, the real-world producer of the traceback you hit). Running the new sanitizer on that schema yields:

```json
"anyOf": [{}, {}, {}]
```

for **every** provider, since `_format_and_dedupe_tools()` has no provider awareness. So OpenAI, Anthropic, and Ollama — which previously received and could honor the constraint — now get a vacuous schema. The failure mode is concrete: the model can emit a targetless `HassTurnOn` call that the advertised schema permits but HA's voluptuous validation rejects at runtime (`required key not provided`), turning a schema-level guardrail into a runtime tool error. The PR description's "No-op for OpenAI/Anthropic" holds for the *object-variant filtering* leg, but not for this strip leg.

The contrast with `_ensure_array_items()` is worth calling out: that one is safe to run unconditionally because it's purely **additive** (injects `items`). This function is **subtractive**, so it needs to be scoped to the provider that can't accept the input.

**Suggested fix:** thread the provider into `_format_and_dedupe_tools()` and only run `_sanitize_any_of_required()` when the chat provider is Gemini. The provider is already available where tools are assembled (`opts.get(CONF_CHAT_MODEL_PROVIDER)` is read in the same node that calls `_format_and_dedupe_tools`, around `graph.py:1603`/`:1624`).

## P2 — drop emptied variants instead of leaving `anyOf: [{}, {}]` on the wire

After stripping, the emptied `{}` variants (and the now-vacuous `anyOf` key) survive into the declaration. langchain_google_genai passes `anyOf` through to the gapic proto without recursing, so Gemini receives `any_of` entries with `TYPE_UNSPECIFIED` — your live test says Gemini tolerates it today, but none of the new tests pin that post-sanitize payload shape, and it's cheap to be strictly safer: when a variant becomes `{}`, drop it; when all variants drop, remove the `anyOf` key entirely.

Relatedly — since for Gemini the constraint genuinely can't be expressed in the proto schema, consider preserving the *semantics* instead of just deleting them: append a hint to the tool/property description, e.g. `"At least one of: name, area, floor is required."`. Gemini reads descriptions even though it can't validate the constraint, and it meaningfully cuts the odds of the model emitting a targetless call that then fails at runtime.

## P2 — hardening: malformed `required` crashes the whole tool-selection node

`graph.py:1259` (`[r for r in sanitized["required"] if r in props]`) raises `TypeError` if a variant's `required` is `null` or a non-list, and `_format_and_dedupe_tools()` has no per-tool exception boundary — one malformed schema from a custom serializer would take down every tool for the turn. Current producers (voluptuous_openapi, pydantic) always emit well-typed lists, so this is hardening rather than a blocker: an `isinstance(sanitized["required"], list)` guard (and `isinstance(schema["anyOf"], list)` before iterating) covers it.

## P3 — nits

- **Implicit-object variants lose `required` entirely.** A variant with `properties` but no explicit `"type": "object"` hits the strip branch instead of the filter branch. Latent today (both producers emit explicit `type: object` on object variants), but `sanitized.get("type") == "object" or "properties" in sanitized` would future-proof the filter leg.
- **Test count:** the diff adds 8 tests, not 9 as the description says.
- **Coverage gaps worth one test each:** a non-dict `anyOf` variant passing through untouched (the `isinstance` guard's false path), an `anyOf` nested inside an `anyOf` variant (the recursion works — I probed it — but nothing pins it), and an assertion on the final emitted shape once emptied-variant dropping lands.

## Summary

1. Gate the sanitizer on the Gemini provider (P1 — the blocker).
2. Drop emptied variants / the vacuous `anyOf` key; optionally move the constraint into the description for Gemini (P2).
3. `isinstance` guards on `anyOf`/`required` (P2, hardening).
4. Test-count fix + the three coverage nits (P3).

The Gemini-side behavior itself is correct and field-verified — this is scoping, not rework. Happy to re-review quickly once the gate is in.

goruck:
@hruba202 — closing the loop here: your fix shipped today in **[v3.26.1](https://github.com/goruck/home-generative-agent/releases/tag/v3.26.1)**, via #536.

Rather than have you rework the branch after the review, I folded your work into #536 directly — the root-cause analysis, the sanitizer approach, and your repro tests are all yours, and the PR and changelog credit you for them. What #536 adds on top is exactly the follow-up items from the review:

- The strip is now **gated on the Gemini provider** (threaded in from tool retrieval), so OpenAI/Anthropic/Ollama keep receiving the at-least-one-target `anyOf` branches they already honour.
- Variants emptied by the strip are dropped, and a fully emptied `anyOf` key removed, so no `TYPE_UNSPECIFIED` entries reach the proto.
- What the strip removes is restated in the schema description ("At least one of: name, area, floor is required."), so Gemini still reads the constraint it can no longer be made to validate.
- Implicit-object variants (`properties` without an explicit `type`) are filtered as objects rather than stripped wholesale, and `isinstance` guards keep one malformed schema from taking down tool selection for the whole turn.

Your three-branch repro is in the suite, alongside a regression test pinning that non-Gemini providers keep the constraint — 20 sanitizer tests in all.

If you get a chance to field-validate on your Gemini install after updating, two things would confirm it end to end:

1. Turns that offer the HA intent tools (`HassTurnOn` and siblings) no longer fail with the `any_of[N].required: only allowed for OBJECT type` 400.
2. Gemini still supplies a target (name/area/floor) when calling those tools — the constraint now travels in the description rather than the schema, so a targetless call that HA then rejects at runtime is the regression to watch for.

Thanks again for the report, the patch, and the repro — this one was entirely your diagnosis.

PR #537, opened 2026-08-09
fix: localize critical-action PIN flow user-facing strings (en/cs)

hruba202:
The v3.26.0 critical-action PIN gate returns fixed user-facing text directly from Python (tools.py, graph.py) as tool results / ToolMessage content. These strings never pass through strings.json/translations, so a Czech-configured instance still sees the PIN flow entirely in English (automation confirmation prompts, PIN validation errors, pending-action expiry messages, etc).

Add agent/pin_messages.py: a small static en/cs lookup table keyed by message id, resolved from hass.config.language with a fallback to English. A static table (not an LLM call) is used because these messages are deterministic, not model-authored, unlike sentinel_response_language/vlm_response_language.

Route every PIN-flow user-facing string in tools.py and _maybe_require_pin_for_automation/_critical_action_guard in graph.py through pin_msg(). LOGGER.warning() diagnostic lines are left in English intentionally, since those are for the HA log, not the chat.


goruck:
Thanks @hruba202 — the approach here is right, and consistent with how this repo handles localization: the PIN-flow strings are deterministic, so a static table (not the LLM) is exactly the mechanism we want, and `hass.config.language` matches what `conversation.py` already uses as the language source.

I pushed a hardening commit (6aabf71) on top rather than bouncing the PR back:

- **The `en` table was missing `automation_pending_store_unavailable`** (it existed only in `cs`), so an English instance hitting the no-pending-store path replied with the raw message id instead of a sentence. GitHub CI doesn't run pytest, but the existing test (`test_add_automation_pin.py`) caught it locally. Restored the original English copy byte-identically.
- **Routed the remaining PIN-flow strings** that were still hardcoded English: missing tool name, HA LLM API unavailable, execute/install failures, invalid pending automation, and the raced not-found-or-expired path in `_execute_pending_automation`. I wrote the Czech for these in your style — please sanity-check my translations.
- **Added `tests/.../test_pin_messages.py`**: en/cs key-set parity (this test alone would have caught the missing key), per-key placeholder parity, format-time brace validation, language resolution (`None` hass, unsupported language, `cs-CZ` region suffix, unknown key), and an end-to-end check that an incorrect PIN on a Czech instance produces the Czech message.
- **Documented the behavior** in `docs/configuration.md` (server language, en/cs, English fallback, logs stay English).

Two deliberate scope choices: `"Configuration not found. Please check your setup."` stays English — it's shared by eight tools outside the PIN flow and fires before `hass` is resolved; and the table is en/cs only while UI translations cover en/cs/ru/tr — ru/tr fall back to English until a native speaker wants to add them (adding a language is just a new `_MESSAGES` key; the parity tests enforce completeness).

Full suite is green locally (2168 tests), lint and pyright clean.

goruck:
Merged and released as [v3.27.0](https://github.com/goruck/home-generative-agent/releases/tag/v3.27.0) — thanks @hruba202 for finding the gap and for the clean static-table approach.

Two-point field validation whenever you get a chance on your instance:

1. **Czech PIN flow end to end** — a critical direct command and a critical automation ("odemkni dveře, když přijdu domů") should now produce the requires-PIN prompt, wrong-PIN error, and expired-confirmation message in Czech. Note the agent's final chat reply is still model prose; what's localized is the deterministic tool-result text it works from, so an occasional English paraphrase from the model itself isn't a bug — raw English *tool strings* leaking through would be.
2. **Sanity-check my Czech** in the strings 6aabf71 added (install/execute failures, missing tool name, LLM API unavailable, invalid pending automation) — wording fixes gladly taken.

Issue #481, opened 2026-07-15
Sentinel trigger queue flooded by non-security camera/person entities (e.g. person_location map cameras) — entity exclusions for event-driven triggering


hruba202:
### System Health details

Just wanted to say thanks; I'm testing it out right now. Great work.

### Checklist

- [x] I have enabled debug logging for my installation.
- [x] I have filled out the issue template to the best of my ability.
- [x] This issue only contains 1 issue (if you have multiple issues, open one issue for each issue).
- [x] This issue is not a duplicate issue of any [previous issues](https://github.com/ludeeus/integration_blueprint/issues?q=is%3Aissue+label%3A%22Bug%22+)..

### Describe the issue

tanks

### Reproduction steps

1.thanks
2.
3.
...

### Debug logs

```text
thanks
```

hruba202:
Summary

Sentinel's event-driven trigger scheduler treats every entity in the camera and person domains as security-relevant (see _DOMAIN_TO_ANOMALY_TYPE in sentinel/engine.py). This works fine for actual security cameras, but breaks down for setups that use camera.* entities for non-security purposes — most notably the built-in person_location platform, which exposes static map snapshots (e.g. Google Static Maps / Mapbox) as camera entities.

Environment


HGA version: 3.14.31
Home Assistant: 2026.x (Proxmox VM, HAOS)
Relevant entities: 14x camera.* entities from the person_location platform (one per household member per map provider — Google/Mapbox/Radar, plus combined multi-person maps), plus at least one additional non-security camera.* entity used for dashboard screenshots (camera.anniroza_screenshot)


Problem

Each person_location camera entity has a state: template that evaluates to stationary / motion_detected based on the tracked person's direction attribute. This means the entity's state changes on essentially every location update (GPS/BLE presence recalculation) — often multiple times per minute per entity, across all 14 entities.

Because _anomaly_type_for_state() maps any camera domain state change unconditionally to camera_entry_unsecured, and any person domain state change to open_entry_while_away, this floods SentinelTriggerScheduler's bounded queue (QUEUE_MAX_SIZE = 10). Since these are classified as security-critical anomaly types, the queue fills entirely with them and legitimate future triggers get dropped:

Trigger queue full (all security-critical); dropping incoming trigger type=camera_entry_unsecured.
Trigger queue full (all security-critical); dropping incoming trigger type=unknown_person_camera_no_home.
Trigger queue full (all security-critical); dropping incoming trigger type=open_entry_while_away.

[...]

PR #544, opened 2026-08-12
Fix Sentinel entity exclusions ignoring phantom lock/door evidence

hruba202:
Two related problems reported by a user whose ESPHome touch-panel template locks (which just mirror a real lock elsewhere, e.g. so a `panel_chodba` or `playercontrol` panel can display/control the front gate) were being treated by Sentinel as real, independently unsecured entry points:

1. `camera_entry_unsecured` never honored `sentinel_rule_entity_exclusions` for the "unsecured" lock/door entity, only for the camera. That rule's `AnomalyFinding.triggering_entities` contains just the camera -- the unsecured entity lives in `evidence` -- so the engine's post-hoc `_filter_excluded_findings` (which only inspects `triggering_entities`) could never suppress it, no matter what was excluded in config. `CameraEntryUnsecuredRule` now accepts an `is_entity_excluded` predicate and filters `unsecured_by_area` (both same-area entities and `camera_entry_links` cross-area entities) before they ever become evidence. `SentinelEngine` wires its existing `_entity_excluded_for_type` through.

2. Configuring exclusions required hand-typing raw JSON (`{"*": ["lock.x", "binary_sensor.y"]}`) into a single-line text field -- easy to get wrong, and not discoverable. The common case (exclude an entity from every rule) now has a proper multi-select entity picker in the Sentinel options flow; the JSON field remains for the rare per-rule-only override and is relabeled "Advanced".

Adds regression tests for the evidence-only exclusion gap in both the same-area and `camera_entry_links` cross-area paths.


goruck:
@hruba202 — closing the loop: this shipped today in **[v3.29.0](https://github.com/goruck/home-generative-agent/releases/tag/v3.29.0)**. Your commits are in as-is; I pushed hardening on top of your branch rather than sending you back around the review loop.

The diagnosis was the hard part and it was yours. `camera_entry_unsecured` putting only the camera in `triggering_entities` while the unsecured lock lives in `evidence` — and therefore `_filter_excluded_findings` structurally never being able to see the entity the user excluded — is not something you find by reading the config code. Your fix is also correctly surgical: I checked that a phantom and a genuinely unlocked door in the same area still produces one finding with only the real lock in evidence, rather than suppressing the whole thing.

What I added on top:

- **A glob under `"*"` would have made the Sentinel settings form permanently unsubmittable.** The split between the picker and the JSON field keyed on the map key, so a stored `{"*": ["camera.map_*"]}` got fed to the `EntitySelector` — which validates through `cv.entity_id_or_uuid` and rejects globs. The sharp edge is that voluptuous validates an `Optional` field's `default` even when that field is absent from the submission, so one stored glob rejected *every* save of the entire form, not just that field. And since `_rule_entity_exclusions_advanced_json` stripped `"*"`, the offending value was invisible in both fields — no way to fix it from the UI. That config isn't hypothetical: `docs/sentinel.md` recommends exactly it for the person_location map cameras from #481. The split is now on **entry shape** rather than map key — literal entity IDs to the picker, globs stay in the advanced JSON — which round-trips losslessly and lets an already-affected install heal itself the next time the form renders, so no storage migration was needed.
- **An excluded entity could still supply the camera's activity timestamp.** When a camera advertises no `last_activity`, the rule falls back to the most recent `last_changed` among nearby sensors, and that scan didn't check exclusions. If someone's phantom is a door/window `binary_sensor` on the panel rather than a template lock, excluding it dropped it from evidence but its chatter kept the camera looking active — so the alert kept firing and the exclusion looked like it did nothing. Exclusions now apply there too, and in-rule suppressions are counted in the rule's debug skip tally instead of a silent `continue` on a high-severity finding.
- Picker entries are validated with `valid_exclusion_entry` like JSON entries are — the entity picker will hand back a bare registry UUID, which has no dot and so gets silently discarded by the engine at load.
- The picker's help text now carries the trade-off the docs already described: an excluded entity stops waking Sentinel, so anomalies involving it are caught on the next poll rather than instantly.

Tests: an engine-level one pinning that the exclusion predicate is actually wired into the rule (deleting that one keyword argument previously left the whole suite green), the merge and round-trip cases for the picker/JSON split, and the two-entity case where only the phantom is excluded.

Your evidence-vs-`triggering_entities` observation generalizes, by the way — `alarm_disarmed_during_external_threat` has the same shape, with secondary panels in `evidence` only. That's filed as a P2 for a shared scrub pass rather than another per-rule callback.

If you get a chance to field-validate after updating, three things would confirm it end to end:

1. Your panel's template lock stops producing `camera_entry_unsecured` alerts once it's in the new picker.
2. A **real** unsecured entry in the same area still alerts. The fix drops only the excluded entity from evidence, not the finding — so a genuinely open door going quiet is the regression to watch for, and the one I'd most want to hear about.
3. The picker round-trips: pick an entity, save, reopen the form (it should show in the picker and *not* in the Advanced JSON box), deselect it, save, reopen — it should be gone rather than coming back.

Thanks for this one. The report, the root cause, and the UI idea are all yours.

PR #545, opened 2026-08-12
Fix OpenAI 400 on tools with a top-level anyOf/oneOf/allOf schema

hruba202:
```
openai.BadRequestError: Invalid schema for function 'HassStartTimer':
schema must have type 'object' and not have 'oneOf'/'anyOf'/'allOf'/
'enum'/'const'/'not' at the top level.
```

HA's built-in `HassStartTimer` intent accepts its `duration` parameter as either a plain string or a structured hours/minutes/seconds object. voluptuous_openapi's conversion of that `vol.Any(...)` emits a top-level `{"anyOf": [...]}` for the tool's `parameters` schema (no `type` key at all). `_format_and_dedupe_tools` then hit its `elif not parameters.get("type")` branch and injected `"type": "object"` -- but left the `anyOf` key sitting right next to it, which is exactly the shape OpenAI's function-calling validator rejects.

Adds `_flatten_top_level_union`, which unions the `properties` of every oneOf/anyOf/allOf variant into a single permissive object schema before the type/properties normalization runs. Required fields are dropped since they differ per variant.

Rebased onto v3.28.0 (main); resolved a placement conflict with the new Gemini `anyOf`-required sanitization helpers added alongside in that release -- both sets of helper functions now live side by side ahead of `_format_and_dedupe_tools`, no logic overlap.

### DIFF
--- a/custom_components/home_generative_agent/agent/graph.py
+++ b/custom_components/home_generative_agent/agent/graph.py
@@ -1335,6 +1335,41 @@ def _sanitize_any_of_required(schema: dict[str, Any]) -> dict[str, Any]:
     return schema
 
 
+_TOP_LEVEL_UNION_KEYS = ("oneOf", "anyOf", "allOf")
+
+
+def _flatten_top_level_union(schema: dict[str, Any]) -> dict[str, Any]:
+    """
+    Flatten a top-level oneOf/anyOf/allOf into a single object schema.
+
+    OpenAI's function-calling validator rejects oneOf/anyOf/allOf/enum/
+    const/not at the top level of ``parameters`` ("schema must have type
+    'object' and not have 'oneOf'/'anyOf'/'allOf'/'enum'/'const'/'not' at
+    the top level"). HA's voluptuous->JSON-schema conversion emits exactly
+    this shape for services with multiple parameter forms, e.g.
+    HassStartTimer's ``duration``, which accepts either a plain string or
+    a structured hours/minutes/seconds object.
+
+    Union the ``properties`` from every variant into one permissive object
+    schema so OpenAI accepts it. Required fields are dropped since they
+    differ per variant and the union is inherently optional-only.
+    """
+    union_key = next((k for k in _TOP_LEVEL_UNION_KEYS if k in schema), None)
+    if union_key is None:
+        return schema
+    variants = schema[union_key]
+    if not isinstance(variants, list):
+        return schema
+    merged_properties: dict[str, Any] = {}
+    for variant in variants:
+        if isinstance(variant, dict):
+            merged_properties.update(variant.get("properties", {}))
+    schema = {k: v for k, v in schema.items() if k not in _TOP_LEVEL_UNION_KEYS}
+    schema["type"] = "object"
+    schema["properties"] = {**merged_properties, **schema.get("properties", {})}
+    return schema
+
+
 def _format_and_dedupe_tools(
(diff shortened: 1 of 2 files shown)


goruck:
@hruba202 — thank you for this one. The find was real and the diagnosis of *where* OpenAI chokes was exactly right; it shipped today as **v3.28.1** via #546, which supersedes this PR (you're credited as co-author on the core commit). Closing this PR in favor of that release — here's the full story, because what we found while verifying is genuinely interesting:

**Why your setup hit it and most didn't.** On HA 2026.8.1 the schema that OpenAI rejects isn't a `duration` string/object union — `HassStartTimer` has no `duration` slot. It's `vol.Required(vol.Any("hours", "minutes", "seconds"))`, which voluptuous_openapi emits as a top-level `anyOf` of bare `{"required": [...]}` variants (same for `HassDecreaseTimer`). Crucially, HA only exposes the timer tools when the request carries a `device_id` that has registered a timer handler (`async_device_supports_timers`) — i.e. a timer-capable voice satellite. So every **voice** request from such a device 400'd while the chat box worked, which is why the bug looked intermittent and why few OpenAI users ever saw it.

**Why we reworked the fix rather than merging as-is.** The unconditional flatten ran for every provider, and HA's at-least-one constraints live in exactly the shape it strips — so it deleted them for Anthropic/Ollama too, and because it ran *before* the Gemini-gated sanitizer, it also starved the description-hint restatement from #536. The two guard tests from that episode (`test_format_and_dedupe_tools_non_gemini_keeps_any_of_required`, `..._gemini_rewrites_ha_target_constraint`) fail on this branch. Same class as the #527 → #536 story: subtractive normalizers need a provider gate.

**What shipped in v3.28.1:** `_flatten_top_level_union` gated to `openai`/`openai_compatible`; variant `required` constraints are restated in the tool description ("At least one of: hours, minutes, seconds is required.") via the same helper the Gemini path uses; other providers get the schema byte-for-byte untouched; malformed variant fields are skipped instead of raising; plus regression tests against the real HA 2026.8.1 emitted schema.

**Field validation ask (2 points), when you get a chance on v3.28.1:**
1. With OpenAI as provider, speak to your timer-capable satellite: "set a timer for 5 minutes" — the 400 should be gone and the timer should actually start (and ring).
2. The regression to watch: a timer call arriving with **no** hours/minutes/seconds field on any provider — that would mean the description hint isn't landing and the model is emitting an empty call that HA then rejects.

Thanks again — this also flushed out a latent gap we're tracking separately (mixed-provider fallback chains bind one tool list to all members), so the report paid for itself twice.

hruba202:
Thanks for the deep dive and for reworking this properly instead of just merging the quick fix — the root-cause writeup (top-level `anyOf` from `vol.Any("hours","minutes","seconds")`, gated only behind timer-capable satellites) explains exactly why this only ever showed up on voice and never in the chat box.

Makes sense to gate `_flatten_top_level_union` to openai/openai_compatible only — good catch on the collision with the Gemini-gated sanitizer and the #536 description-hint restatement. Nice callback to the #527 → #536 pattern too.

I'll pull v3.28.1 and run both checks:
1. "Set a timer for 5 minutes" via my timer-capable satellite on OpenAI — confirming no 400 and the timer actually starts/rings.
2. Watching logs for any timer call landing with no hours/minutes/seconds field, to make sure the description-hint restatement is actually steering the model.

Will report back here with results. Thanks again!

PR #559, opened 2026-08-18
v3.30.7 fix: spurious listener warning, and the orphaned video analyzer behind it

hruba202:
## Problem

The one-time `EVENT_HOMEASSISTANT_STARTED` listener that starts the video analyzer self-unsubscribes once it fires (that's what `async_listen_once` does). The remove function it returns was also wrapped in `entry.async_on_unload(...)`. If the config entry reloads/unloads *after* the event already fired, HA calls that same remove function a second time, which tries to remove an already-removed job from the listener list, raising a caught `ValueError` and logging:

```
Unable to remove unknown job listener ... ValueError: list.remove(x): x not in list
```

Benign, but noisy - shows up in the log every time the entry reloads post-startup.

## Fix

Guard the removal with a `fired` flag so the remove function is only invoked if the listener hasn't already fired and self-unsubscribed.

Note: the neighbouring `_start_sentinel` / `_start_discovery` listeners don't wrap `async_on_unload` around their `async_listen_once` call at all, so they aren't affected by this.

---

## Maintainer hardening

Thanks @hruba202 - the diagnosis is right, and it's worse than a noisy log. Verified against the installed HA 2026.8.1: the remover returned by `async_listen_once` is not idempotent, and `_async_remove_listener` logs the second call via `_LOGGER.exception`, so it's an ERROR with a traceback rather than a warning.

Two things were added on top of the original commit. Both came out of the pre-landing review; the second was found by the Codex adversarial pass *after* the first was already in.

### 1. The `fired` flag was written on the wrong thread (`d0741ed`)

`_start_video_analyzer` had no `@callback`, so `HassJob` infers `Executor`:

```
PR version job_type : HassJobType.Executor
@callback version   : HassJobType.Callback
```

`_OneTimeListener` unsubscribes on the loop thread and *then* hands the body to a worker thread (`core.py:349-362` → `932-938`). Between those two points the listener is already gone but `fired` is still `False`, so unload calls the stale remover and logs the exact error the flag was added to prevent.

`@callback` on both closures puts the whole body in the same loop tick as the unsubscribe. `call_soon_threadsafe` is no longer needed and would only reopen the window by deferring the start a tick, so `start()` is called directly - matching the `hass.is_running` branch above it.

### 2. The cancel cannot close the orphan window on its own (`da00220`)

`async_unload_entry` calls `video_analyzer.stop()` - which returns immediately at `video_analyzer.py:2950` when the analyzer never started - and then awaits five more times before HA runs the entry's on-unload callbacks. `EVENT_HOMEASSISTANT_STARTED` firing in any of those windows starts the analyzer and *correctly* leaves the cancel with nothing to do. HA then deletes `runtime_data` around a live capture loop whose retention deque has deletion power over the replacement entry's snapshot files - the exact hazard `be3f85c` set out to prevent.

`stop()` now sets a one-way latch before its not-started early return (the entry that never started is exactly the one whose deferred start is still armed) and `start()` checks it first. The latch sits downstream of every ordering; the cancel does not. A `VideoAnalyzer` belongs to exactly one setup - constructed in `async_setup_entry`, stopped only in `async_unload_entry` - so there is no legitimate stop-then-start cycle to break.

### Tests

`tests/custom_components/home_generative_agent/test_deferred_start_listener.py`, 4 tests on the existing `test_fallback_setup` harness. Two of them fail on the code before each fix:

- `test_started_listener_runs_inline_on_the_event_loop` - `assert [] == [<_MainThread...>]` without `@callback`
- `test_stop_latches_the_analyzer_against_a_later_deferred_start` - `AttributeError: 'MockConfigEntry' object has no attribute 'runtime_data'` without the latch, which is the production symptom

`make lint` clean, 2560 tests pass.

### Deferred

You were right that the neighbours are a separate matter, but they're worse off than this one: `_start_sentinel`, `_start_discovery`, and `_start_baseline` have no `async_on_unload` at all, so a reload before HA finishes starting leaves them running the orphaned engines alongside the replacement entry. Separately the `EVENT_HOMEASSISTANT_STOP` listener leaks one closure per reload, and `openai_http_client` is only ever closed from inside it. All three are filed under a new `## Config Entry Lifecycle` section in `TODOS.md` for a focused follow-up - including the note that copying just the cancel to those three would copy this PR's orphan window along with it.


goruck:
Merged and released as v3.30.7. Thanks @hruba202 — the diagnosis was right, and chasing it turned up two more orderings behind the log line: the flag was being written on an executor thread, and separately the cancel can't close the orphan window at all, because HA runs an entry's on-unload callbacks only after `async_unload_entry` returns. Both are fixed on top of your commit, with tests that fail without each one.

Two things to watch on your box:

1. Reload the integration a few times after HA is fully up — the `Unable to remove unknown job listener` traceback should be gone entirely.
2. If you ever reload while HA is still starting, watch for duplicate camera analysis or snapshots disappearing sooner than your retention setting allows. That was the real hazard; it should now be impossible.

PR #561, opened 2026-08-20
v3.30.9 fix: warn when a pipeline setting silently disarms the critical-action PIN

goruck:
> **This PR changed direction after field testing.** Its original premise was wrong and is documented below rather than deleted, because the wrong version was pushed and reviewed.

## Summary

Reported symptom: with **Prefer handling commands locally** enabled, "lock the garage door" and "unlock the garage door" execute without a critical-action PIN prompt, and a follow-up "unlock it" replies "Unlocking" without unlocking anything.

**This is designed Home Assistant behavior, not an HGA bug, and no integration can intercept it.** When a pipeline has `prefer_local_intents` on, HA matches the sentence against its own built-in intents and executes on a match *before* calling the conversation agent. The agent never sees the turn, so its PIN cannot gate it. **The fix for the user is to turn that option off.**

`ConversationEntityFeature.CONTROL` does not change this. It has exactly one consumer in all of HA (`assist_pipeline/pipeline.py:1259`), which installs `_async_local_fallback_intent_filter` — and `default_agent.async_handle_intents` treats that filter as a **reject list**:

```python
# default_agent.py:1458
result = await self.async_recognize_intent(user_input, strict_intents_only=True)
if not isinstance(result, RecognizeResult) or (
    intent_filter is not None and intent_filter(result)
):
    return None          # True => do NOT handle locally
```

The filter matches only `HassGetState` and media search, so *those two* go to the agent and every control command stays local regardless of the flag.

### What this PR now does

- **Adds a repair issue** (`core/pipeline_guard.py`) raised when a critical-action PIN is configured while a pipeline routing to this agent has `prefer_local_intents` on. It names the offending pipelines, tells the user which setting to turn off, and clears when resolved. This is visibility, not interception — the point is that the gap was completely silent.
- **Restores the original docs guidance.** `docs/configuration.md` already said to disable "Prefer handling commands locally" for the PIN to work; that was correct, and an earlier commit on this branch wrongly replaced it with the opposite. Restored with the reason, plus the two bypasses that setting does not cover: sentence triggers (`pipeline.py:1236` runs them ahead of any agent) and anything that isn't the conversation agent.
- **Keeps one real fix**: `active_llm_api_ids()` gives the three readers of `CONF_LLM_HASS_API` a single "absent means Assist" default. The options flow deletes the key when no API is selected, so the entity was claiming no entity control while the Assist API was active. Consistency fix, **not** a security fix.
- **Drops the PIN clause on the CONTROL flag** — it was built on the wrong premise and does nothing for the PIN.

### Correction history

The first three commits claimed this closed a PIN bypass. That was wrong. I inferred the filter's semantics from its name and the `pipeline.py` comment ("we filter out some sentences that are interfering with LLM operation") and never read the call site that *consumes* `intent_filter`. Two rounds of Codex review ran inside that same frame and agreed, so cross-model agreement did not catch it — field testing did. Commits are kept rather than rewritten so the record is auditable.

## Test Coverage

```
CODE PATHS
[+] core/pipeline_guard.py                                       [NEW]
  ├── [★★★] PIN on + local intents on + routes to us -> issue, pipeline named
  ├── [★★★] PIN off                        -> issue cleared
  ├── [★★★] local intents off              -> issue cleared
  ├── [★★★] pipeline routes to another agent -> ignored
  ├── [★★★] several conflicting pipelines  -> all named, sorted
  ├── [★★★] issue id scoped per config entry
  ├── [★★★] assist_pipeline absent         -> silent, never raises
  └── [★★★] assist_pipeline installed but not set up -> silent

[~] conversation.py :: CONTROL feature gate
  ├── [★★★] absent key -> CONTROL set (capability/tools agreement)
  ├── [★★ ] configured API -> CONTROL set
  └── [★★★] stored [] + PIN on -> NO CONTROL (pins the corrected contract)

[~] agent/helpers.py :: active_llm_api_ids()   6 paths, unchanged from earlier revision

COVERAGE: 17/17 paths tested (100%)  |  GAPS: 0
```

Suite: **2591 passed**. `make lint` clean. `pyright` 2 errors, both pre-existing in untouched code. Translation parity (30 tests) green — `strings.json`, `en.json`, and `cs.json` all carry the new `issues` block.

⚠️ **The Czech translation of the repair issue needs a native check** before merge.

## Pre-Landing Review

Earlier revision: 2 findings, both auto-fixed (module bound under two names; test docstrings describing an unreachable state). Current revision adds no new findings.

## Adversarial Review

Codex raised a P1 on the earlier revision (explicit `[]` from the v5→v6 migration was a second unguarded state) and it was correct *within the wrong premise* — I had asserted `[]` was unreachable and it is not. That finding is now moot: neither spelling was ever protected by the flag.

Structured review gate on the earlier revision: **PASS**, one P2 accepted (the docs claimed local handling was restricted to state reads; media search also actuates). That P2 was pointing at the right function and I drew the wrong conclusion from it.

## Design Review

No frontend files changed — design review skipped.

## Eval Results

No prompt-related files changed — evals skipped.

## Scope Drift

Scope Check: **CLEAN**.

## TODOS

P2 under **Agent**: "no LLM APIs" has two storage spellings that behave oppositely — deselecting in the options flow pops the key (silently restoring Assist), while the v5→v6 migration writes `[]` (which really disables everything). Independent of this PR's premise and still valid.

## Known gaps

- **Sentence triggers still bypass the PIN** and the repair issue does not detect them. `pipeline.py:1236` runs them ahead of any conversation agent.
- **The repair issue re-evaluates on setup and reload only.** Changing the pipeline setting alone does not reload the config entry, so the issue clears on the next reload or restart rather than instantly.
- **Nothing here makes voice-operated locks safe.** The only reliable control is to stop exposing them to Assist; the docs now say so.

## Test plan

- [x] Full suite (2591 tests, 0 failures)
- [x] `make lint` clean, translation parity green
- [ ] On 192.168.1.240: with a PIN set and "Prefer handling commands locally" **on**, confirm a repair notice appears under Settings → System → Repairs naming the pipeline
- [ ] Turn the option **off**, reload, confirm the notice clears and "lock the playroom door lock" now prompts for the PIN
- [ ] Native check of the Czech strings


goruck:
Merged and released as [v3.30.9](https://github.com/goruck/home-generative-agent/releases/tag/v3.30.9). Field-validated on a live install before merge:

- With a PIN set and **Prefer handling commands locally** on, the repair notice appears naming the pipeline.
- With the option off and the entry reloaded, the notice clears and `lock the garage door lock` prompts for the PIN.

@hruba202 — one small ask if you have a minute. This adds a Repairs notice with new `issues` strings, and `cs.json` needed full key parity so I wrote the Czech myself. It should at least be *actionable*: the setting it tells you to turn off is named in both Czech and English, because "Prefer handling commands locally" is a frontend string I couldn't verify the canonical Czech for from the backend package. But it would benefit from a native read for idiom. The key is `issues.pin_bypassed_by_local_intents` in `custom_components/home_generative_agent/translations/cs.json`.

For the record, since the early commits on this branch are wrong and are kept in history rather than rewritten: this PR started out claiming to *fix* a PIN bypass by setting `ConversationEntityFeature.CONTROL`. That was based on a misreading. `CONTROL` has exactly one consumer in Home Assistant, and the filter it installs is a **reject** list — `default_agent.async_handle_intents` returns `None` when the filter matches, and the filter matches only `HassGetState` and media search. So those two go to the agent and every control command, `HassTurnOn`/`HassTurnOff` on a lock included, stays local regardless of the flag.

Locally-handled lock commands bypassing an agent-side PIN is therefore designed Home Assistant behavior that no integration can intercept, and the pre-existing documentation telling you to disable that option was correct all along. What shipped is a warning so the misconfiguration stops being silent, plus a restored and expanded docs section. It does not make voice-operated locks safe — unexposing the entity from Assist is still the only control that also covers sentence triggers and non-agent paths.

PR #565, opened 2026-08-21
Localize Sentinel notification title, subtitle, and type labels

hruba202:
## Summary

Follow-up on the review discussion in #523, where the deterministic notification bodies were kept English-only by design, but title/subtitle translation was flagged as a genuinely open piece of work:

> if you want to take the title/subtitle translation, that one's genuinely open and I'd review it

This PR localizes the parts of a Sentinel notification that are *not* LLM output -- the fixed Python string literals in `notifier.py` that #523 explicitly left out of scope.

## What's added

`sentinel/notifier_messages.py`: a small `en`/`cs` message table with `notif_msg(hass, key, **kwargs)`.

Language is resolved from `hass.config.language` (base language code, e.g. `cs-CZ` -> `cs`), **not** `CONF_SENTINEL_RESPONSE_LANGUAGE`. That option is a free-text LLM prompt fragment ("Czech", "please respond in German"), not a language code, so it can't be used as a lookup key. Using `hass.config.language` also means these strings localize automatically on a Czech-configured instance even if `sentinel_response_language` is left unset -- which matches how `pin_messages.py` already resolves language for pinned-message translation.

## What's localized

- Severity titles (`Security Alert` / `Home Alert` / `Home Update`)
- Finding-type labels (`_KNOWN_TYPE_LABELS`) -- including `appliance_power_duration`, which was missing from the curated label dict entirely and fell through to the raw-slug fallback
- Subtitle templates (`_build_subtitle`)
- Burst-batch summary title/message
- Daily digest title/message, including the severity words in the summary (`2 high, 1 low`)
- Permanent-snooze confirmation title/message
- The `_fallback_message` / `_persistent_message` fallback strings

## What's deliberately out of scope

The deterministic per-template mobile message bodies (`_alarm_disarmed_mobile_message`, `_appliance_power_duration_mobile_message`, `_baseline_deviation_mobile_message`, `_entity_staleness_mobile_message`, `_alarm_disarmed_open_entry_mobile_message`) stay English-only, matching the #523 discussion: they render exact figures, units, and entity names that a translation could blur, and for the security-relevant ones (camera name, disarm time, entry name, call to action) losing precision matters more than the language it's in. Only the chrome around them is localized here.

## Adding a language

Add one dict to `_MESSAGES` in `notifier_messages.py` with the same keys as `"en"`. Missing keys in a non-English language fall back to English automatically, so a partial translation degrades gracefully rather than crashing.

## Testing

Verified with `ruff check` / `ruff format` (clean) and `ast.parse` on both files. I wasn't able to run the `pytest-homeassistant-custom-component` suite in the sandbox I drafted this in (Python 3.12 there vs. `homeassistant==2026.8.1`'s `>=3.14.2` requirement), so I'd appreciate CI/your review flagging anything the existing `test_sentinel_notifier.py` suite catches that I should account for -- happy to iterate.

Manually verified on my own HA instance (Czech-configured) against real Sentinel findings, including the previously-missing `appliance_power_duration` type label.


goruck:
Thanks @hruba202 — this is the right design and I'm taking it. Keying off `hass.config.language` instead of `sentinel_response_language` was exactly the correct call (that option is a free-text prompt fragment, not a language code), the en/cs key parity was perfect, and you even caught that `appliance_power_duration` was missing from the curated label table entirely. The #531 boundary survives intact: deterministic security bodies stay English, only the chrome localizes.

You asked the review to catch what your sandbox couldn't run — it did. **29 of 126 notifier tests crashed**: `getattr(hass.config, "language", None)` guards `language` but dereferences `hass.config` unguarded, and the suite's minimal `DummyHass` has no `config`. Since `async_notify` is the last hop of the alert pipeline, that pattern would also turn any degraded runtime state into a lost security notification instead of an English one. CI runs no pytest, so all four checks were green over it.

I've pushed the fix plus hardening as da3b22b on your branch (maintainer-edit, same as #537):

- `_resolve_language` guards the `config` attribute itself
- `notif_msg` now keeps the docstring's no-crash promise at format time too: a bad placeholder degrades cs → en → unformatted instead of raising, and an unknown/data-derived message id is returned verbatim, never treated as a format template (a stored digest severity that collided with a real message id would otherwise surface unrelated copy)
- the English fragments inside otherwise-Czech strings ("Entry", "Sensor", "Unknown entity") became `fallback_*` keys
- the daily digest survives malformed stored records (`finding: null`, non-dict, `severity: null`)
- the persistent fallback renders the localized severity word ("závažnost střední", not "závažnost high")
- 25 new tests: a parity/contract suite for `notifier_messages.py` (key parity, placeholders-vs-call-site, degrade paths), Czech end-to-end through `async_notify`/batch/digest/snooze, and a pin that security bodies stay English even on a cs-configured instance

2616 tests pass, ruff and pyright clean.

Three things I'd like your input on, as the Czech speaker:

1. **Check my Czech.** I added `Vstup` / `Senzor` / `Neznámá entita` for the three fallback words — please correct them if they read wrong in context.
2. **Batch pluralization.** `"{count} novinek z domova"` is the 5+ genitive, but the batch typically flushes 1–4 items, so the common renderings ("1 novinek", "2 novinek") are wrong Czech. A count-agnostic phrasing like `"Novinky z domova ({count}): {type_summary}."` would sidestep declension entirely — your call on the wording; happy to take it as a follow-up commit here.
3. **Scope check on two gaps** (fine as follow-up PRs if you'd rather keep this one focused):
   - The action buttons stay English — including `Confirm`/`Cancel` under the now-Czech permanent-snooze prompt, which is a destructive action guided by English verbs. Both review models flagged this one.
   - Built-in types not in the curated table (`vehicle_detected_near_camera_home`, `pet_detected_at_night_no_occupancy`, `unknown_person_camera_no_home`, `phone_battery_low_at_night_home`, `camera_missing_snapshot_night_home`, …) still render English slug-prettified subtitles on a Czech instance. Pre-existing behavior, but if you want to fill in cs labels for them, this table is now the natural home.

Once merged and released, a quick field validation on your instance would be great: (a) Czech titles/subtitles/digest on real findings, (b) the security message bodies (alarm-disarmed family) still arriving in exact English, and (c) no raw `type_*`/message ids anywhere.

goruck:
Merged and released as [v3.31.0](https://github.com/goruck/home-generative-agent/releases/tag/v3.31.0) — thanks @hruba202, this is a genuinely nice piece of work and it's now running on my own install.

Rather than hold the merge on the open review questions, I shipped it as-is; please look them over post-merge and open an issue (or a follow-up PR) for anything that needs correcting:

1. **My Czech needs a native check** — I added `Vstup` / `Senzor` / `Neznámá entita` for the fallback fragments inside Czech templates. If any of these read wrong, that's a one-line fix.
2. **Batch plural** — cs `batch_message` uses "{count} novinek", which I believe is wrong for counts 1–4 ("novinka"/"novinky"/"novinek"). The batch typically flushes 1–4 items, so this shows often. A count-agnostic phrasing like "Novinky z domova ({count}): …" would also work — your call on wording.
3. **Action buttons still English** — "Confirm"/"Cancel" under the Czech permanent-snooze prompt (plus "Ask Agent" etc.). Flagged in review as the top remaining gap since Confirm/Cancel guides a destructive action; deliberately left out of this PR's scope.
4. **Uncurated type labels** — several built-in types (`camera_missing_snapshot_night_home`, `unknown_person_camera_no_home`, `phone_battery_low_at_night_home`, `vehicle_detected_near_camera_home`, `pet_detected_at_night_no_occupancy`) aren't in the label table yet, so they render as prettified English slugs on Czech installs. Needs native-Czech labels.

Field validation when you get a chance: with HA set to Czech, (a) a normal finding shows a Czech title/subtitle while security bodies stay English, (b) a burst of 4+ low-severity findings produces a Czech batch summary, and (c) the daily digest arrives in Czech.

Issue #571, opened 2026-08-25
Sentinel: evidence-less low_battery candidates key subject=unknown, never dedup against evidenced duplicates


hruba202:
### Summary

Two pending Sentinel discovery candidates for the *same* low-battery sensor never dedup against each other, so both stay visible in the Proposals card indefinitely:

- `low_battery_sensor_0xffffaa67127301f8` — Type: `unspecified`
- `v1|subject=sensor|predicate=low_battery|night=any|home=any|scope=...` — Type: `maintenance`

### Root cause

`candidate_semantic_key()` in `discovery_semantic.py` resolves the `low_battery` predicate from the `candidate_id` slug alone (`_has_low_battery_signal(text, slug_text)`), independent of whether any `entities[...]` evidence_paths actually resolve to a concrete `sensor.*` id.

When `_battery_sensor_entity_ids(entity_ids)` comes back empty (the candidate cites no machine-parseable entity evidence — just a free-form `candidate_id` like `low_battery_sensor_<id>` plus prose), the function still returns:

```
v1|subject=unknown|predicate=low_battery|night=any|home=any|scope=any|entities=
```

instead of `None`. That's worse than no key at all, in two directions:

1. **Under-merge (the reported symptom):** it can never equal a properly-evidenced duplicate's key (e.g. `v1|subject=sensor|predicate=low_battery|...|entities=sensor.xxx_battery`), so the two candidates about the same physical sensor never dedup.
2. **Over-merge:** it's *constant* — every evidence-less low-battery candidate, regardless of which sensor it's actually about, produces the exact same string, so two candidates about *different* sensors would incorrectly collide if both happened to lack resolvable evidence_paths.

### Repro

```python
from custom_components.home_generative_agent.sentinel.discovery_semantic import candidate_semantic_key

no_evidence = {
    "candidate_id": "low_battery_sensor_0xffffaa67127301f8",
    "title": "Nízká úroveň baterie senzoru 0xffffaa67127301f8",
    "summary": "Baterie senzoru 0xffffaa67127301f8 klesla pod doporučenou hranici.",
}
evidenced = {
    "candidate_id": "v1|subject=sensor|predicate=low_battery|night=any|home=any|scope=any",
    "title": "Nízká úroveň baterie senzoru 0xffffaa67127301f8",
    "summary": "Upozornit, když kapacita baterie senzoru 0xffffaa67127301f8 klesne pod definovaný práh.",
[...]

PR #572, opened 2026-08-25
v3.31.3 fix: evidence-less low_battery candidates key None, not subject=unknown

hruba202:
Refs #571. (Maintainer note: keeping #571 open — see the review comment below. The key-function bug this fixes is real, but the two-card symptom in the screenshot survives, as the Known limitation section already says.)

### Problem

Two pending Sentinel discovery candidates for the same low-battery sensor never dedup against each other (screenshot in the issue). One candidate has a free-form `candidate_id` (`low_battery_sensor_<id>`) and no machine-parseable `entities[...]` evidence — just prose naming the sensor — the other is a properly-evidenced, structured candidate for the same sensor.

### Root cause

`candidate_semantic_key()`'s low_battery branch resolves the predicate from the `candidate_id` slug alone, independent of whether any entity evidence actually resolves. When `_battery_sensor_entity_ids()` comes back empty, it still emitted:

```
v1|subject=unknown|predicate=low_battery|night=any|home=any|scope=any|entities=
```

That key is *constant* regardless of which sensor the candidate is actually about, so it:
- **under-merges** the reported case (never matches a properly-evidenced duplicate about the same sensor), and
- **over-merges** the unreported case (two candidates about *different* sensors would incorrectly collide if both lacked resolvable evidence).

### Fix

When no battery entity id resolves, revert the predicate back to `"unknown"` so the function's existing `subject == "unknown" and predicate == "unknown"` check returns `None`, routing the candidate through `_candidate_identity_hash` (title+summary hash) — the same fallback other null-key candidates already use (the Bug 2 fix elsewhere in `discovery_engine.py`).

### Testing

- Added `test_zero_evidence_low_battery_candidate_keys_none` and `test_zero_evidence_low_battery_candidates_do_not_force_collide`.
- Manually verified against the existing low_battery test scenarios in `test_discovery_semantic.py` (lock+battery, night-context collapse, weak-wording, device-named, motion-named, battery-arm needle tests) via an isolated import of `discovery_semantic.py` — all unaffected, since they all supply resolvable `entities[...]` evidence_paths.
- `ruff check` passes on both changed files.

I wasn't able to spin up the full `homeassistant`-pinned test env in my sandbox (dependency resolution for the exact pinned HA/langchain versions timed out), so I verified `candidate_semantic_key` in isolation rather than running `pytest` end-to-end for this repo — happy to address anything CI turns up.

### Known limitation (left as follow-up)

This stops the "generic key" failure mode, but doesn't *guarantee* the originally-reported pair now dedups — the evidence-less candidate falls back to `_candidate_identity_hash`, which only collides on near-identical title+summary wording, and the two cards in the screenshot are worded differently. A more complete fix would tighten the discovery LLM prompt/schema so a low-battery candidate naming a specific sensor can never omit a matching `entities[entity_id=...]` evidence path in the first place. Flagging in case that's worth a separate issue.

---

### Summary

hruba202's diagnosis and destination were both right: the low_battery leg emitted a constant `v1|subject=unknown|predicate=low_battery|…|entities=` key whenever `_battery_sensor_entity_ids()` resolved nothing, and `_candidate_identity_hash` is the correct fallback. Review found the mechanism had two holes, and both were fixed on the branch (c257272):

**The `return None` never fired for window-worded candidates.** Setting `predicate = "unknown"` and relying on the trailing `subject == "unknown" and predicate == "unknown"` guard is action-at-a-distance — the subject back-fill (`"window"`/`"windows"` prose → `entry_window`) runs in between and rescued the candidate straight back onto a constant key. Window and door contacts are the battery-powered devices this leg sees most. The leg returns `None` directly now.

**The `else` was unscoped, and that part was a regression.** `_battery_sensor_entity_ids` only ever returns `sensor.*` IDs, so it is also empty for a battery candidate citing a lock, an alarm panel, or a `binary_sensor` contact — the normal way Zigbee/Z-Wave devices are cited. Those keep a resolved subject, never reach the `None` guard, and were silently losing their predicate: `"Low battery on the alarm panel"` keyed identically to any unrelated predicate-less candidate about the same panel, and `_filter_novel_candidates` dropped one for the other in LLM emission order. The `None` return is gated on `subject == "unknown"`, so those shapes keep their pre-existing keying.

Both were found independently by the maintainability, testing, and Claude adversarial passes and by Codex.

### Test Coverage

Ship-stage coverage audit of the changed leg (2d8c947). The `None` path is reached by what the evidence paths *resolve to*, not by whether the key exists, so it now covers index-based paths (`entities[31].state`), derived-only paths, non-entity paths, and evidence in domains the subject chain ignores (camera, device_tracker, light, cover). Base parity is pinned for the two remaining subject legs (`entry_window`, `lock`) and for the `subject=sensor` shapes where `_battery_sensor_entity_ids` returns `[]` for a reason other than "no evidence".

The consumer side was previously unverified — asserting the key is `None` says nothing about whether the engine's `key or _candidate_identity_hash(...)` fallback actually dedups, and the derived-only and entity-text-mismatch guards run *before* the dedup check. Both dedup legs are now driven through `_filter_novel_candidates`, plus the over-merge half of #571 (two different sensors both survive as novel).

Tests: 2722 → 2738. Coverage of the changed leg: complete.

### Pre-Landing Review

Two P1s found and fixed (above). Ruff and pyright clean; pyright unchanged at its 5 pre-existing repo-wide errors.

### Adversarial Review

Claude adversarial + Codex, on both the review diff and the final shipped diff. The final Codex pass returned five findings: three were already filed in TODOS.md this cycle, and two produced changes in 4cca086 —

- The CHANGELOG over-claimed "one card per sensor". The identity hash is over `title + summary`, so two identically-worded proposals about different sensors still merge, and a re-proposal carrying a live reading still mints a second card. Both limits are named in the entry now.
- `test_low_battery_sensor_subject_without_battery_id_keeps_base_keying` pinned an exact key without recording that the key is a known mirror mismatch (a lone `sensor.*_power` with battery prose keys `predicate=low_battery` while the normalizer registers `baseline_deviation`). The test exists to pin base parity, not to bless the key; its docstring says so now.

### Scope Drift

Scope Check: CLEAN. Intent — stop evidence-less low_battery candidates keying a constant string. Delivered exactly that, scoped to the shape that resolves nothing, plus tests and follow-up tracking.

### Plan Completion

No plan file — contributor PR.

### TODOS

No existing items completed. Four new P2 follow-ups filed, all surfaced by reviewing this change:

- The null-key path persists the model's own supplied `semantic_key` (`enriched = dict(candidate)`, computed key written back only `if key:`), and `_collect_existing_keys` then trusts it. Pre-existing; this change enlarges the population that reaches it.
- Identity-hash dedup cannot collapse re-proposals whose prose carries a live reading.
- Stored discovery keys are never migrated when the key chain changes.
- The `unlocked` / `open` / `unavailable` / `disarmed` legs sit above the battery leg and still have the constant evidence-less key shape this PR removed from it. `"Battery sensor unavailable"` with no evidence keys a constant every similarly-worded candidate shares.

### Issue #571

Deliberately **not** auto-closed — `Fixes` was changed to `Refs`. The key-function bug is fixed, but the two-card symptom in the screenshot survives: the evidence-less card falls back to an identity hash, which can never equal the evidenced card's semantic key. hruba202 called this out himself under "Known limitation", and his suggested prompt/schema fix is the durable answer.

### Test plan

- [x] Full suite: 2738 passed
- [x] `ruff check` + `ruff format --check` clean across 174 files
- [x] `pyright`: no new errors
- [ ] Field validation by @hruba202 (2 points, asked in the review comment)


goruck:
Thanks @hruba202 — the diagnosis is right and the write-up made this easy to verify. The constant `v1|subject=unknown|predicate=low_battery|...|entities=` key really does over-merge distinct sensors and under-merge against evidenced duplicates, and routing to `_candidate_identity_hash` is the correct destination.

I pushed a hardening commit (c257272) rather than asking for changes, because reviewing it turned up two things that only show up downstream of the key function. Both were found independently by three review passes plus Codex, and both are reproduced below against your branch as it stood.

**1. The `return None` never fires for window-worded candidates.**

The sentinel value isn't inert — two fixups run between the assignment and the guard it relies on:

```python
if subject == "unknown" and _contains_any(text, ("window", "windows")):
    subject = "entry_window"
```

That runs *after* the battery leg, so:

```
"Low battery on the bedroom window sensor" → v1|subject=entry_window|predicate=unknown|night=any|home=any|scope=any|entities=
"Low battery on the kitchen window sensor" → v1|subject=entry_window|predicate=unknown|night=any|home=any|scope=any|entities=
```

Byte-identical for two different physical sensors — the same over-merge, just relabeled from `predicate=low_battery` to `predicate=unknown`. Window and door contacts are the battery-powered devices this leg sees most, so it isn't an exotic shape. Both of your tests use Czech prose with no `window` token, which is why the suite stayed green. The leg now returns `None` directly instead of falling through.

**2. The `else` wasn't scoped to the evidence-less shape.**

`_battery_sensor_entity_ids` only ever returns `sensor.*` IDs, so it comes back empty for any battery candidate citing a lock, an alarm panel, a `binary_sensor` contact, or two ambiguous non-battery sensors — the normal way Zigbee/Z-Wave devices get cited. Those candidates keep a resolved subject, so they never reach the `None` guard; they just quietly lost their predicate:

```
"Low battery on the alarm panel"          → v1|subject=alarm|predicate=unknown|...|entities=alarm_control_panel.house
"Alarm panel reports implausible values"  → v1|subject=alarm|predicate=unknown|...|entities=alarm_control_panel.house
```

Identical. Driven through the real `_filter_novel_candidates`, one drops the other as `batch_duplicate` / `existing_semantic_key`, and *which* one survives depends on LLM emission order or on what's already in the 200-record history window. A dead alarm-panel battery losing to an unrelated benign candidate about the same panel is a worse outcome than the bug being fixed. On `main` these key `predicate=low_battery` and both are kept.

The fix is one line different from yours — `elif subject == "unknown": return None` — which closes both: returning directly skips the back-fill, and the `subject == "unknown"` gate keeps the evidenced shapes at their existing keying. Their known mirror mismatches are already recorded in TODOS.md; I'd rather leave those than trade them for a new collision.

**Tests.** Added the window shape, plus the three non-`sensor.*` evidence shapes pinned to `predicate=low_battery`. I also reworked `test_zero_evidence_low_battery_candidates_do_not_force_collide` — its name promises non-collision but the body asserted `is None` twice, which would pass even if both candidates collapsed onto one hash. It now compares the *effective* dedup key (`candidate_semantic_key(c) or _candidate_identity_hash(c)`), which is what the engine actually dedups on. 2722 tests pass, ruff and pyright clean.

**On the env.** No need to apologise for not running the suite — `make venv && make testdeps` is the whole story if you want it locally, but CI plus this review is enough. Your isolated verification of `candidate_semantic_key` was the right call and it held up.

**On your Known limitation — you were right, and I've kept #571 open because of it.** I dropped the `Fixes` keyword from the description so merging won't auto-close it. Your instinct about the prompt/schema is the real fix and I've filed it, along with two related things this review turned up:

- The identity-hash fallback hashes `title\0summary`, and candidate prose carries live readings. Three re-proposals of one sensor with a drifting percentage now produce three cards where the old constant key produced one. So for the evidence-less shape this trades over-merge for flood — worth knowing, and the reason your prompt/schema suggestion is the durable answer rather than a nicety.
- Stored discovery keys are never migrated. `_collect_existing_keys` prefers a record's stored `semantic_key` over recomputing it, so after this lands, previously-suppressed evidence-less battery topics — including **rejected** ones — reappear once as fresh cards before settling. Self-healing after a cycle, and there's precedent (the v3.25.0 migration did the same), but it'll go in the changelog.
- Separately, Codex caught that on the null-key path the engine keeps whatever `semantic_key` the *model* supplied (`enriched = dict(candidate)`, and the computed key is only written back under `if key:`). Pre-existing, but this PR enlarges the population that reaches it. Filed as P2.

All three are in TODOS.md in the hardening commit.

**Field validation, when you get a chance** — two things I can't check from here:

1. On your box with the two pending low-battery cards for `0xffffaa67127301f8`: after this ships, confirm the evidence-less card no longer suppresses or gets suppressed by unrelated candidates, and tell me whether the pair still sits there side by side (expected — that's #571 staying open) or whether the count grows across cycles (that's the flood above, and it'd bump the prompt/schema fix up the list).
2. If you have a battery-powered window or door contact that generates candidates, confirm those cards now behave independently per sensor instead of collapsing onto one.

Issue #524, opened 2026-07-31
hga-proposals-card.js: away/home/night badge detection breaks for non-English candidate text


hruba202:
## Is your feature request related to a problem? Please describe.

`hga-proposals-card.js` infers away/home/night context for a discovery
candidate by regex-matching English keywords against its `title`/`summary`
text (`isAway`, `isHome`, `hasNight` in the badge-kind resolver, e.g. around
`www/hga-proposals-card.js:272-291` and again at `:378-389`):

```js
const hasNight =
  evidencePaths.includes("derived.is_night") || text.includes("night");
const isAway =
  /\b(?:away|no one home|nobody home|empty|unoccupied|no occupants|without occupants)\b/.test(
    text
  );
const isHome =
  /\b(?:someone home|occupied|home|present|occupants|residents)\b/.test(
    text
  ) || evidencePaths.includes("derived.anyone_home");
```

This is only a fallback — `evidencePaths.includes("derived.is_night")` /
`derived.anyone_home` are checked first and are language-independent — but
`isAway` currently has **no** `evidence_paths`-based path at all, so it's
100% dependent on the English substring/regex match.

With the `sentinel_response_language` option added in #523, a candidate's
`title`/`summary` can be written in the user's configured language (e.g.
Czech). None of the English keywords above will ever appear in translated
text, so:

- the away badge (`isAway`) never fires for translated candidates — it has
  no structured fallback,
- the home/night badges silently fall back to whatever `evidence_paths`
  provides, which is better but not always present (e.g. `hasNight`'s text
  fallback and `entryKind`'s `/\bwindows?\b/`/`/\bdoors?\b/` matching at
  `:288-292` have the same gap).

The card already has precedent for handling localized text gracefully:
`entryTextRe`/the locale-named-entity-ID fallback added in #504 promotes
unclassified `binary_sensor.`/`cover.` entities to entry candidates when the
text doesn't carry an English "door"/"window" token, specifically because
locale-named entities (e.g. Czech "okno") don't carry an English token
either. The away/home/night detection doesn't have an equivalent fallback.

## Describe the solution you'd like

Make `isAway`/`isHome`/`hasNight`/`entryKind` primarily structured
(`evidence_paths`-driven) rather than text-driven, with the English regex
kept only as a legacy fallback for candidates generated before
`sentinel_response_language` existed:

- discovery/triage prompts already require `evidence_paths` entries
  referencing concrete entity IDs — worth checking whether a small,
[...]

Issue #522, opened 2026-07-30
[Rule] Nízká baterie zámku dveří


hruba202:
### Rule name

n_zk_baterie_z_mku_dve

### Summary

Baterie senzoru sensor.zamek_vrata_baterie klesla pod nastavenou hranici kritické kapacity.

### Motivation and safety impact

This proposal is currently marked unsupported in HGA and needs a deterministic template.
Candidate ID: zamek_vrata_baterie_low_battery
Observed pattern: threshold_breach

### Required snapshot evidence

- entities[sensor.zamek_vrata_baterie].state

### Deterministic detection logic

1) Evaluate only snapshot fields listed in required evidence.
2) Trigger when the candidate condition is true.
3) Return no findings when any required condition is missing.

### Evidence mapping

- evidence.path_1 <- entities[sensor.zamek_vrata_baterie].state

### Severity

medium

### Confidence (0.0-1.0)

0.7

### Sensitive action?

- [ ] This rule is sensitive and must require confirmation for any action.

### Suggested actions (semantic only)

- close_entry

### Suppression/cooldown expectations

Use default per-type cooldown (30 min).

### Test cases

- Trigger: Baterie senzoru sensor.zamek_vrata_baterie klesla pod nastavenou hranici kritické kapacity.
- Non-trigger: same snapshot context but with condition absent.

goruck:
Released in **v3.23.1** — thanks @hruba202, this one was a genuinely interesting report: your candidate was unsupported for three stacked reasons, and it's the first field proof that discovery prose isn't always English.

**What was fixed:**
1. The evidence format `entities[sensor.zamek_vrata_baterie].state` (bare-bracket) wasn't parsed — now it is, on every surface (normalizer, dedup keys, hallucination guard, proposals card).
2. The prose being Czech meant no English keyword could match. The low-battery signal is now also read from the machine-generated candidate ID (`zamek_vrata_baterie_low_battery`), matched on whole tokens and scoped strictly to battery routing.
3. `baterie` doesn't contain the English `battery` token, so the sensor was never collected — a single unambiguous `sensor.*` evidence entity is now promoted as the battery sensor.

Bonus fixes that fell out of review: Czech decimal commas parse correctly ("pod 20,5 %" → 20.5, previously 5.0), and a threshold embedded in the candidate ID (`…_battery_below_10`) registers as 10% instead of the 40% default.

[...]

Issue #516, opened 2026-07-29
[Rule] Motion detected at night while away


hruba202:
### Rule name

motion_detected_at_night_while_away

### Summary

Trigger when a motion sensor reports ON at night and nobody is home.

### Motivation and safety impact

This proposal is currently marked unsupported in HGA and needs a deterministic template.
Candidate ID: v1|subject=motion_sensor|predicate=state_event|night=1|home=0|scope=any|entities=binary_sensor.xiao_esp32_c5_espectre_motion
Observed pattern: state_event

### Required snapshot evidence

- derived.is_night
- derived.anyone_home
- entities[entity_ids contains binary_sensor.xiao_esp32_c5_espectre_motion].state

### Deterministic detection logic

1) Evaluate only snapshot fields listed in required evidence.
2) Trigger when the candidate condition is true.
3) Return no findings when any required condition is missing.

### Evidence mapping

- evidence.path_1 <- derived.is_night
- evidence.path_2 <- derived.anyone_home
- evidence.path_3 <- entities[entity_ids contains binary_sensor.xiao_esp32_c5_espectre_motion].state

### Severity

medium

### Confidence (0.0-1.0)

0.8

### Sensitive action?

- [ ] This rule is sensitive and must require confirmation for any action.

### Suggested actions (semantic only)

- close_entry

### Suppression/cooldown expectations

Use default per-type cooldown (30 min).

### Test cases

- Trigger: Trigger when a motion sensor reports ON at night and nobody is home.
- Non-trigger: same snapshot context but with condition absent.

goruck:
Shipped in [v3.22.0](https://github.com/goruck/home-generative-agent/releases/tag/v3.22.0) (PR #520) — thanks @hruba202, this is now a supported deterministic template. Your proposal was well-specified enough to implement almost verbatim.

What you should see after updating:

- Your candidate (`v1|subject=motion_sensor|…|entities=binary_sensor.xiao_esp32_c5_espectre_motion`) normalizes into an approvable rule instead of carrying the unsupported badge — approve it from the proposals card as usual.
- The rule fires when any of its motion sensors reports ON at night while nobody is home. It is suppressed during the departure/arrival grace window, so tripping the sensor yourself on the way out at night shouldn't page you.
- If a sensor in the rule is later renamed or removed, the remaining sensors keep alerting (missing IDs are recorded in the finding evidence) — the rule only goes silent if *none* of its sensors resolve.

[...]

PR #573, opened 2026-08-25
v3.32.1 fix: null-key candidate semantic_key leak + stable battery-slug identity (#571 follow-ups)

hruba202:
Refs #571. Two related follow-ups Codex's adversarial review surfaced during the #572 review (both now in TODOS.md under "Discovery"):

### 1. Drop model-supplied `semantic_key` on null-key candidates

`DISCOVERY_OUTPUT_SCHEMA` lets the model supply an optional `semantic_key` field. In `_filter_novel_candidates`, when `candidate_semantic_key()` returns `None`, `enriched = dict(candidate)` left that field untouched and it was persisted into the discovery record.

`_collect_existing_keys` prefers a stored `semantic_key` over recomputation:

```python
key = str(candidate.get("semantic_key", "")) or candidate_semantic_key(candidate)
```

so an LLM-declared string on a null-key candidate could suppress an unrelated, differently-keyed real proposal as `existing_semantic_key` for as long as it stays in the 200-record discovery-history window.

**Fix:** drop the field explicitly on the null-key path so a null-key candidate can only ever be recalled via `_candidate_identity_hash`, matching what `_collect_existing_keys` actually falls back to for it.

### 2. Stable identity for evidence-less low-battery candidates

`_candidate_identity_hash` fell back to hashing `title+summary` for every null-key candidate. LLM candidate prose routinely embeds the current reading ("Baterie klesla na 12 %" then "11 %" next cycle), so the prose hash alone made every cycle's re-proposal of the same evidence-less battery sensor a new pending card — the exact symptom the #571 screenshot showed (an evidence-less card and an evidenced card for the same sensor sitting side by side, plus the same-sensor re-proposal drift noted in the maintainer's own follow-up comment on #571).

**Fix:** `battery_slug_device_token()` (new, in `discovery_semantic.py`) extracts a device/sensor token from the candidate's `candidate_id` slug (e.g. `low_battery_sensor_0xffffaa67127301f8` → `0xffffaa67127301f8`), gated on the same low-battery signal the battery leg in `candidate_semantic_key()` already checks. `_candidate_identity_hash` hashes that token instead of the drifting prose when one resolves.

Deliberately conservative: after stripping known topic words (`low`, `battery`, `sensor`, `level`, `capacity`, ...), exactly one token must remain — zero means no device-identifying text to anchor on, more than one is ambiguous about which token names the device. Both cases fall back to the prose hash rather than guess, since a false match would merge two genuinely different sensors (worse than the drift this fixes).

### Testing

- New tests in both `test_discovery_engine_dedupe.py` and `test_discovery_semantic.py`, including:
  - a poisoned-`semantic_key` candidate no longer carries it through `_filter_novel_candidates`
  - three re-proposals of one sensor with a drifting battery percentage now collapse to one identity hash
  - two different zero-evidence battery sensors still don't collide
  - a non-battery null-key candidate's hashing is unaffected
  - an ambiguous multi-token slug falls back to the prose hash rather than guessing
- Verified in isolation (same constraint as #572 — couldn't get the full pinned `homeassistant`/`langchain` test env running in my sandbox) against `discovery_semantic.py` directly, and mirrored every existing + new `test_discovery_engine_dedupe.py`/`test_discovery_semantic.py` assertion by hand in that harness — all green, including the existing upstream tests these changes touch (`test_window_worded_zero_evidence_low_battery_candidate_keys_none`, `test_zero_evidence_low_battery_candidates_do_not_force_collide`, `test_candidate_identity_hash_is_stable`, `test_candidate_identity_hash_differs_on_content`).
- `ruff check` passes across the whole repo.

### Still open after this

Per the maintainer's own #571 comment and the third TODOS.md item ("Earlier predicate legs still emit constant evidence-less keys"), the `unlocked`/`open`/`unavailable`/`disarmed` legs in `candidate_semantic_key()`'s elif chain have the same constant-key shape the battery leg had — untouched by #572 and untouched here. Each leg needs its own review against the normalizer's branch order (`open`/`unlocked` can legitimately key entity-less for the any-window template, so a blanket guard would break that dedup), so I left it out of this PR rather than rush it. Happy to pick it up separately if useful.

---

## Maintainer ship (v3.32.1)

Appended below @hruba202's description; nothing above this line was changed.

Both diagnoses here are correct and both close the TODOs they target. Three mechanisms needed work, all reproduced locally and independently confirmed by a Codex pass. Full write-up in [this comment](https://github.com/goruck/home-generative-agent/pull/573#issuecomment-5434919596).

**The branch broke an existing green test** — `test_filter_zero_evidence_low_battery_dedups_on_identity_hash`, `assert 2 == 1`. CI here runs HACS, Hassfest, Ruff and claude-review but **no pytest**, so all four checks were green over a real regression, and @hruba202 had said up front he could not get the pinned env running. Worth remembering on the next one.

Hardening in `227f697`:

1. **Identity namespace split.** `_candidate_identity_hash` picked one of two namespaces (device token or prose) while the engine compared a single key, so a re-proposal whose slug drifted never met its twin — the #571 flood symptom on the other axis. `_candidate_identity_keys` now returns a set, matched by intersection. Forcing that relation transitive was tried and reverted: the union of two heuristic equivalences is not transitive, and propagating dropped candidates' keys let A link B to C when B and C shared no key, hiding a distinct card.
2. **Unvalidated device tokens.** `low_battery_sensor_kitchen` and `kitchen_battery_low` both reduce to `kitchen`, merging two real sensors. Six shape tests went in and five were broken by a place label wearing the same grammar (`kitchen1`, `basement1`, `room1234`, `12345678`, `a1234567`). Landed on the `0x` literal marker over a non-null hex body of 8+ characters, and nothing else.
3. **Mirror drift.** The battery gate read `title+summary` while `candidate_semantic_key`'s battery leg reads `_candidate_text_blob` (+ `pattern` + `suggested_type`). Both read the same surface now.

Also: `_collect_existing_keys` always recomputes and never trusts a stored `semantic_key`, which closes the poisoned-history residue and the paired migration gap together, as TODOS.md already prescribed.

**Design rule, now in the comments:** a rejected token costs some re-proposal drift; a wrong one hides a low-battery card the user never sees. Always narrow.

`665ff6a` merges main (v3.32.0) in so the bump starts from the current version, bumps to v3.32.1, and files two residuals in TODOS.md — the English-only slug topic-word list (P2, open with @hruba202 on the thread) and the shared-`0x`-address merge (P3, accepted knowingly).

**#571 stays open**, same reason as after #572: an identity hash can never equal a semantic key, so a rejected evidence-less proposal still returns once it gains evidence. Untouched by this diff.

2848 tests, ruff clean, pyright unchanged (5 pre-existing errors in files this PR does not touch).


goruck:
## Maintainer review

Thanks for this. Both diagnoses are correct and both close the TODOs they target: the null-key path really did persist the model's own `semantic_key`, and the identity hash really was hashing prose that carries a live reading. I have pushed one hardening commit to your branch (`227f697`) rather than sending it back, because the fixes turned out to need a couple of rounds and it was faster to work them through directly. Your two commits are untouched underneath it.

Everything below was reproduced locally and independently confirmed by a Codex pass.

### First, the thing neither of us could see from CI

The branch broke an existing green test:

```
test_filter_zero_evidence_low_battery_dedups_on_identity_hash
>       assert len(filtered) == 1
E       AssertionError: assert 2 == 1
```

`test_discovery_engine_dedupe.py` is 43/43 on `main` and 42/43 here. This is not on you. You said in the description that you could not get the pinned `homeassistant`/`langchain` env running in your sandbox, and CI here runs HACS, Hassfest, Ruff and claude-review but **no pytest at all**, so all four checks were green over a real regression. Worth both of us remembering on the next one: green CI on this repo says nothing about the test suite.

### 1. The two identity namespaces split candidates apart

`_candidate_identity_hash` picked *one* of two namespaces, device token or prose, and `_filter_novel_candidates` compared a single key. So a candidate whose slug resolved a token hashed as `battery-device\0<token>`, while its re-proposal whose slug picked up one extra word fell back to `title\0summary`. The two never met.

That is the same flood symptom #571 reports, just on the other drift axis: the PR stabilises prose drift and introduces slug drift. The failing test above is exactly that case, with `_again` appended to the slug.

Fixed by keeping `_candidate_identity_hash` as the canonical prose hash and adding `_candidate_identity_keys`, which returns *every* key a null-key candidate can be recalled by. Dedup matches on set intersection, so whichever surface stayed stable is always compared.

### 2. Any sole leftover token was trusted as a device identity

`return remaining[0]` had no shape check, so:

```
low_battery_sensor_kitchen  -> kitchen
kitchen_battery_low         -> kitchen
```

Two unrelated sensors in one room collapse into one card, and the second is silently dropped. The `len(remaining) != 1` rule does not prevent the false merge your comment reasons about, it only narrows coverage. Also `low_battery_sensor_98` yielded `98`.

This one took the most work. Six shape tests went in and five got broken by a room label wearing the same grammar:

| Rule | Broken by |
|---|---|
| contains a digit | `kitchen1` |
| token must appear in the prose | `basement1` corroborates itself, and a MAC written `AA:BB:CC:DD:EE:FF` stops matching `aabbccddeeff` |
| at least half digits | `room1234` |
| bare hex-ness | `12345678` |
| hex with an a-f letter | `a1234567` |

The corroboration attempt is the instructive failure: the model authors both the slug and the prose, so repeating a token in one is not independent evidence of the other.

What landed is the `0x` literal marker over a non-null hex body of at least 8 characters, and nothing else. A room or unit label does not carry `0x`, and it is exactly the Zigbee IEEE address shape your #571 screenshot shows. Bare MAC-style ids are deliberately given up: an unrecognised token falls back to the prose hash, which is what `main` already does, so that costs coverage and never correctness.

Your instinct in the PR description was the right one and I have made it the explicit design rule in the comments: a rejected token costs some re-proposal drift, a wrong one hides a low-battery card the user never sees. Always narrow.

### 3. The gate did not mirror the gate it says it mirrors

`battery_slug_device_token` built `text = f"{title} {summary}"`, but `candidate_semantic_key`'s battery leg gates on `_candidate_text_blob`, which is title + summary + **`pattern`** + **`suggested_type`**. Reproduced with a Czech-prose candidate whose only English battery surface is `suggested_type: "low_battery_sensors"`:

```
semantic key            : None
blob-gate (key leg)     : True
narrow-gate (helper)    : False
device token            : None
DRIFT STILL PRESENT     : True
```

The exact class the fix targets, still drifting. Now both read `_candidate_text_blob`. This module has had roughly six of these asymmetries (#516, #518, #522, #524), so it was worth a line in the docstring saying why the surface is shared rather than copied.

### Also in the hardening commit

- `_collect_existing_keys` now **always** recomputes and never trusts a stored `semantic_key`. Your `pop` stops new poisoning, but records already inside the 200-record window still carry one, and the read path trusted it, so the reported suppression survived the fix until those records aged out. Always recomputing is what TODOS.md already prescribed for the paired migration gap, and it closes both at once.
- `_BATTERY_SLUG_TOPIC_WORDS` folds in `_LOW_BATTERY_QUALIFIERS` instead of re-listing `low`/`below`/`under`/`weak`, so adding a qualifier there cannot silently stop it being stripped here.
- Secondary identity keys stay out of `hint_keys`. That set is sorted and truncated to `_MAX_SEMANTIC_KEYS_IN_PROMPT`, and `ident|` sorts ahead of `v1|`, so a second opaque hash per candidate would evict real semantic keys from the prompt for no gain. The LLM cannot compute either hash.
- `_ZERO_EVIDENCE_LOW_BATTERY_CANDIDATE` is gone; the new tests reuse the existing `_ZERO_EVIDENCE_BATTERY_CANDIDATE` about a hundred lines below it.

### Two of my own fixes were wrong

In fairness, since I am marking up yours. I first made batch dedup transitive by feeding dropped duplicates back into `seen_batch`, to kill an order-dependence. That was worse than the problem: the union of two heuristic equivalences is not transitive, so A links B to C when B and C share no key at all, hiding a genuinely distinct card. Reverted, with a comment saying not to retry it. The residual order-sensitivity only bites among candidates with identical prose, which are indistinguishable on the card anyway and which `main` merged unconditionally. The test now pins the guarantee that actually holds, never weaker than the prose hash alone in any ordering, instead of an order-independence the design cannot have.

The six-round token shape saga above was the other one.

### Where this leaves things

2769 tests, `ruff` clean, `pyright` unchanged. All four CI checks green on `227f697`.

**#571 still stays open**, for the same reason as after #572: a rejected evidence-less proposal still fails to suppress the same candidate once it gains evidence, because an identity hash can never equal a semantic key. That is untouched by this diff and is the structural item, not a bug in your work.

You were right to leave the `unlocked`/`open`/`unavailable`/`disarmed` legs out. They need their own pass against the normalizer's branch order, and your read that a blanket guard would break the any-window template's dedup is correct. Happy to take that as a separate PR whenever you want it.

### Two things I would value from your deployment

1. Do your real `candidate_id` slugs carry the `0x` form? Your screenshot shows `low_battery_sensor_0xffffaa67127301f8`, which is why the gate is scoped to it, but you are the one with a live Czech-locale instance producing these. If your model also emits slugs like `nizka_baterie_senzoru_0xffff...`, tell me: the topic-word list is English-only, so a Czech slug never reduces to one token and the improvement will not reach you.
2. After this lands, whether the `0xffffaa67127301f8` sensor stops minting a new pending card each cycle as the percentage drops.

Merging after that, then a release and a bump.

PR #586, opened 2026-08-29
fix: flatten top-level tool-schema unions for Anthropic too (#585)

### LINKED ISSUE #585: Anthropic API rejects Home Assistant Assist tool schemas: input_schema does not support anyOf at the top level
### System Health details

version: 2026.8.3
installation_type: Home Assistant OS
dev: false
hassio: true
docker: true
arch: x86_64
timezone: America/New_York
os_name: Linux
os_version: 6.18.39-haos
supervisor: 2026.08.0
host_os: Home Assistant OS 18.2
docker_version: 29.6.2
python_version: 3.14.6

Home Generative Agent: v3.33.0 (HACS)
Model providers: Anthropic (claude-sonnet-4-6 for conversation and vision, claude-haiku-4-5-20251001 for summarisation) and OpenAI (text-embedding-3-small, embeddings only)
Database: PostgreSQL 17.5 + pgvector via the goruck/addon-postgres-pgvector add-on
Options: llm_hass_api ["assist"], tool_retrieval_limit 5, tool_relevance_threshold 0.15

### Checklist

- [x] I have enabled debug logging for my installation.
- [x] I have filled out the issue template to the best of my ability.
- [x] This issue only contains 1 issue (if you have multiple issues, open one issue for each issue).
- [x] This issue is not a duplicate issue of any [previous issues](https://github.com/ludeeus/integration_blueprint/issues?q=is%3Aissue+label%3A%22Bug%22+)..

### Describe the issue

With Anthropic Claude as the conversation model provider and `llm_hass_api: ["assist"]`, any turn in which the agent retrieves an affected Home Assistant intent tool fails with a 400 before the request ever reaches the model.

```
anthropic.BadRequestError: 400 invalid_request_error —
tools.N.custom.input_schema: input_schema does not support
oneOf, allOf, or anyOf at the top level
```

Raised from `langchain_anthropic/chat_models.py::_astream` → `_handle_anthropic_bad_request`, surfacing as `HomeAssistantError: Model invocation failed` at `agent/graph.py:2042`, and reported to the user as a mid-stream error at `conversation.py:1116`.

## Cause

Home Assistant's Assist intent tools describe their target as "name **or** area **or** floor **or** domain", and the timer intents describe theirs as a union of name/area/duration fields. Both serialise to a top-level `anyOf` in the generated JSON Schema, which the Anthropic Messages API refuses. Validation fails at request time, so the model never sees the turn.

Home Assistant's own built-in Anthropic conversation integration sanitises these schemas before sending them. Home Generative Agent appears to pass them through unmodified.

## Which tools are affected

All 25 `Hass*` intent tools, not just the obvious control ones. I confirmed this by exclusion: excluding 18 action tools left history questions and fall-through commands still failing, because the seven timer tools were still being sent —

```
HassCancelTimer · HassDecreaseTimer · HassIncreaseTimer ·
HassPauseTimer · HassStartTimer · HassTimerStatus · HassUnpauseTimer
```

Excluding all 25 resolved it. `GetLiveContext` and `GetDateTime` are clean and can stay.

## Why this is hard to diagnose

**The 400 names the offending tool by index, not by name** — `tools.2`, `tools.3`, `tools.4` across otherwise identical failures. Because `tool_retrieval_limit` sends only about five semantically-retrieved tools per turn, the index refers to a different tool each time and cannot be mapped back to a name from the error alone.

**It also looks intermittent.** Turns whose retrieved set happens to exclude an affected tool succeed normally, so "which lights are on" and "what time is it" work while "when were the kitchen lights last turned off" fails. That reads as a flaky integration rather than a schema problem.

**And it only appears once vector tool retrieval works.** On the keyword fallback these tools are rarely retrieved, so the schemas are never sent. Fixing retrieval is what exposes it — which will read to users as "fixing the database broke my assistant."

## Workaround in place

All 25 `assist::Hass*` tools excluded via `tool_exclusions`, keeping only `GetLiveContext` and `GetDateTime`. Device control is handled by Home Assistant's local intent engine with `prefer_local_intents: true` on the pipeline.

This works, and commands got faster — but it has a real cost: anything local intents fail to match falls through to the agent, which now has no control tools at all. A slightly misheard command becomes a dead end rather than a polite refusal.

## Suggested fixes, in order of usefulness

1. **Sanitise tool schemas before dispatch when the provider is Anthropic** — flatten a top-level `anyOf`/`oneOf`/`allOf` into a single object schema with the union of properties and no required key, as the core Home Assistant Anthropic integration does.
2. **Failing that, catch the 400 and drop the offending tool** rather than failing the whole turn.
3. **Name the tool in the error.** Mapping the returned `tools.N` index back to the tool name before re-raising would turn a multi-hour diagnosis into a one-line one.

Happy to supply the full traceback, the serialised schema for one offending tool, and the config entry's options dict.

### Reproduction steps

1. Install Home Generative Agent v3.33.0 on Home Assistant 2026.8.3 (HAOS, x86-64).
2. Configure **Anthropic Claude** as the Conversation model provider (`claude-sonnet-4-6`).
3. Add an OpenAI embeddings provider (`text-embedding-3-small`) — required, since Anthropic has no embeddings API.
4. Configure the database via `+ Setup → Advanced` so vector tool retrieval actually works. (On the keyword fallback the bug stays hidden.)
5. Set `llm_hass_api: ["assist"]` and expose at least one light.
6. Ask a question whose retrieved tool set includes any `Hass*` intent tool — for example "when were the kitchen lights last turned off?", or any device command that misses the local intent engine.

**Expected:** the turn completes; the agent answers or acts.

**Actual:** the 400 above. The user hears an error.

### Debug logs

```text
homeassistant.exceptions.HomeAssistantError: Model invocation failed: Error code: 400 - {'type': 'error', 'error': {'type': 'invalid_request_error', 'message': 'tools.3.custom.input_schema: input_schema does not support oneOf, allOf, or anyOf at the top level'}, 'request_id': 'req_011CeVfXLNqZ2Vo1gUV8fqNN'}
anthropic.BadRequestError: Error code: 400 - {'type': 'error', 'error': {'type': 'invalid_request_error', 'message': 'tools.3.custom.input_schema: input_schema does not support oneOf, allOf, or anyOf at the top level'}, 'request_id': 'req_011CeVfXLNqZ2Vo1gUV8fqNN'}
homeassistant.exceptions.HomeAssistantError: Model invocation failed: Error code: 400 - {'type': 'error', 'error': {'type': 'invalid_request_error', 'message': 'tools.2.custom.input_schema: input_schema does not support oneOf, allOf, or anyOf at the top level'}, 'request_id': 'req_011CeVfH8qAhbSxqwoZiTAXF'}
anthropic.BadRequestError: Error code: 400 - {'type': 'error', 'error': {'type': 'invalid_request_error', 'message': 'tools.4.custom.input_schema: input_schema does not support oneOf, allOf, or anyOf at the top level'}, 'request_id': 'req_011CeVakfCT9puoLc7wqgw4K'}

Note the differing tool index (tools.2, tools.3, tools.4) across otherwise identical failures.

Abridged traceback from the system log:

ERROR [custom_components.home_generative_agent.conversation] HomeAssistantError mid-stream; partial content committed.
Traceback (most recent call last):
  File "/config/custom_components/home_generative_agent/agent/graph.py", line 2034, in _invoke_model
    return await model.ainvoke(messages, config)
  File "/usr/local/lib/python3.14/site-packages/langchain_anthropic/chat_models.py", line 1445, in _astream
    _handle_anthropic_bad_request(e)
  File "/usr/local/lib/python3.14/site-packages/anthropic/_base_client.py", line 1940, in request
    raise self._make_status_error_from_response(response) from None
anthropic.BadRequestError: Error code: 400 - invalid_request_error - tools.4.custom.input_schema: input_schema does not support oneOf, allOf, or anyOf at the top level

The above exception was the direct cause of the following exception:

  File "/config/custom_components/home_generative_agent/conversation.py", line 1116, in _async_run_astream
  File "/config/custom_components/home_generative_agent/agent/graph.py", line 2200, in _call_model
  File "/config/custom_components/home_generative_agent/agent/graph.py", line 2066, in _invoke_chat_model_with_sampling_rebind
  File "/config/custom_components/home_generative_agent/agent/graph.py", line 2042, in _invoke_model
    raise HomeAssistantError(msg) from err
homeassistant.exceptions.HomeAssistantError: Model invocation failed
During task with name 'agent' and id 'cbf389ee-f51a-8b17-e101-3411f8381a74'
```

### Diagnostics dump

_No response_

goruck:
Fixes #585. Reported by @a4bell64, whose report identified the cause, the two-line fix, and both fallbacks — this PR implements all three.

## The bug

Anthropic's Messages API rejects `oneOf`/`allOf`/`anyOf` at the top level of a tool's `input_schema`, exactly as OpenAI's validator does:

```
tools.3.custom.input_schema: input_schema does not support oneOf, allOf, or anyOf at the top level
```

Validation runs before the model reads the turn, so one such tool 400s the entire request and the user hears an error.

`_flatten_top_level_union()` has handled this since #546 — but only for `openai` and `openai_compatible`. The gate excluded `anthropic` on the untested assumption that it honours the union. It does not; Home Assistant's own built-in Anthropic integration strips the keys outright before dispatch (`components/anthropic/entity.py::_format_tool`).

## Which tools actually carry the shape

Checked against the installed HA 2026.8.2, converting each intent's real slot schema through `voluptuous_openapi`: **two**, not 25.

| Tool | Top-level union |
|---|---|
| `HassStartTimer` | `anyOf` |
| `HassDecreaseTimer` | `anyOf` |
| `HassCancelTimer`, `HassIncreaseTimer`, `HassPauseTimer`, `HassUnpauseTimer`, `HassTimerStatus`, `HassCancelAllTimers` | none |
| `HassTurnOn`-style target tools | none |

Only `vol.Required(vol.Any("hours", "minutes", "seconds"))` produces it. The bare `vol.Any("name", "area", "floor")` marker key that the target tools use converts to plain optional properties. The report's exclusion experiment is consistent with this: excluding the 18 action tools left the timer group in place, and the timer group is where both offenders live.

## The fix

`_FLATTEN_UNION_PROVIDERS` gains `anthropic`. The dropped constraint survives as the description hint the OpenAI and Gemini paths already use, so the model is still told a duration is required.

Verified end to end against the real 2026.8.2 schemas through `langchain_anthropic.convert_to_anthropic_tool` — which copies `parameters` into `input_schema` verbatim, hence the passthrough:

```
HassStartTimer     union_keys: []  description: "At least one of: hours, minutes, seconds is required."
HassDecreaseTimer  union_keys: []  description: "At least one of: hours, minutes, seconds is required."
```

## The two fallbacks, also from the report

- **A rejected tool no longer takes the turn down.** `_invoke_chat_model_with_schema_recovery()` drops the offending tool, rebinds and retries (bounded at 3 drops). The schema normalisation is the fix for the shapes we know about; this is the net under it for the ones we do not.
- **The tool is named.** Providers identify it only by position, and with `tool_retrieval_limit: 5` that position addresses a different tool each turn — which is what made this a multi-hour diagnosis. The index is now resolved against the bound tool list before it is logged or re-raised. Handles Anthropic's `tools.N.` and Gemini's `function_declarations[N]` forms; anything unrecognised is re-raised untouched.

## For the reporter

The `assist::Hass*` `tool_exclusions` workaround can be removed once this is released.

## Testing

- 2865 tests pass; `make lint` clean; pyright unchanged (5 pre-existing errors on `main`, same 5 here).
- New: Anthropic flatten coverage (target schema + real `HassStartTimer` end to end), rejection-index parsing (Anthropic/Gemini forms, cause-chain walk, unrelated-error and absurd-index guards), and three recovery tests — drop-and-retry against a real `bind_tools` chain, tool named when retries run out, unrelated errors passed through without a rebind.
- The two `union_capable_keeps_*` guard tests now cover `ollama`/unset only; Anthropic moved to the flatten side.

https://claude.ai/code/session_01LhU3C8AkecjNhiGFmv7Fvh

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,14 @@
 
 All notable changes to this project will be documented in this file.
 
+## [3.33.2] - 2026-08-28
+
+### Fixed
+
+- **Claude no longer fails a turn because of how Home Assistant describes its timer commands.** Two of Assist's built-in commands — start a timer and shorten a timer — say "give me hours, or minutes, or seconds" in a form Anthropic's API refuses to accept, and it refuses the whole request rather than the one command, so any turn that happened to offer Claude one of those two tools died with `input_schema does not support oneOf, allOf, or anyOf at the top level` and the user heard an error. The requirement is now restated in the tool's own description before the request is sent, exactly as is already done for OpenAI and Gemini, so Claude still knows a duration is needed. **If you excluded the `Hass*` tools to work around this, you can restore them.** Reported by [@a4bell64](https://github.com/a4bell64) ([#585](https://github.com/goruck/home-generative-agent/issues/585)).
+- **One tool a provider refuses to accept no longer takes the whole reply down with it.** Providers validate every tool before reading the conversation, so a single schema they dislike fails the turn. When that happens the offending tool is now dropped and the turn is retried without it — a lost tool instead of a lost answer.
+- **The log now names the tool a provider rejected.** Providers identify it only by its position in the request ("tools.3"), and because the set of tools sent changes from turn to turn that position points at something different each time. The name is now resolved and written to the log, so the tool to exclude is stated outright instead of having to be deduced.
+
 ## [3.33.1] - 2026-08-28
 
 ### Changed
--- a/custom_components/home_generative_agent/agent/graph.py
+++ b/custom_components/home_generative_agent/agent/graph.py
@@ -1425,9 +1425,13 @@ _TOP_LEVEL_UNION_KEYS = ("oneOf", "anyOf", "allOf")
 
 # Providers whose function-calling validator rejects oneOf/anyOf/allOf at the
 # top level of `parameters`. openai_compatible is included because those
-# endpoints proxy or emulate OpenAI's validation; the schema loss is recovered
-# via the description hint either way.
-_FLATTEN_UNION_PROVIDERS = ("openai", "openai_compatible")
+# endpoints proxy or emulate OpenAI's validation. Anthropic's Messages API
+# rejects the same shape on its own account ("tools.N.custom.input_schema:
+# input_schema does not support oneOf, allOf, or anyOf at the top level",
+# issue #585) — Home Assistant's built-in Anthropic integration strips those
+# keys outright before dispatch. The schema loss is recovered via the
+# description hint either way.
+_FLATTEN_UNION_PROVIDERS = ("anthropic", "openai", "openai_compatible")
 
 
 def _merge_conjunctive_required(
@@ -1460,12 +1464,15 @@ def _flatten_top_level_union(schema: dict[str, Any]) -> dict[str, Any]:
     OpenAI's function-calling validator rejects oneOf/anyOf/allOf/enum/
     const/not at the top level of ``parameters`` ("schema must have type
     'object' and not have 'oneOf'/'anyOf'/'allOf'/'enum'/'const'/'not' at
-    the top level"). On HA 2026.8.1 exactly two assist tools carry that
-    shape — HassStartTimer and HassDecreaseTimer, whose
+    the top level"), and Anthropic's Messages API rejects the union keys
+    the same way ("tools.N.custom.input_schema: input_schema does not
+    support oneOf, allOf, or anyOf at the top level"). On HA 2026.8.2
+    exactly two assist tools carry that shape — HassStartTimer and
+    HassDecreaseTimer, whose
     ``vol.Required(vol.Any("hours", "minutes", "seconds"))`` slot converts
     to a top-level ``anyOf`` of bare ``{"required": [...]}`` variants — and
     both are exposed only for timer-capable satellite devices, so the 400
-    surfaces only on voice requests from such a device.
+    surfaces only on turns whose tool set includes one of them.
 
     Every variant's ``properties`` are unioned into one object schema
     (pre-existing top-level properties win).  ``anyOf``/``oneOf`` variant
@@ -1543,10 +1550,10 @@ def _format_and_dedupe_tools(
         if not isinstance(parameters, dict):
             parameters = {"type": "object", "properties": {}}
         else:
-            # OpenAI rejects oneOf/anyOf/allOf at the top level of
-            # `parameters` (HassStartTimer / HassDecreaseTimer). Subtractive,
-            # so gated — Anthropic/Ollama honour the union and keep it, and
-            # the Gemini pass below needs the anyOf intact to restate it.
+            # OpenAI and Anthropic both reject oneOf/anyOf/allOf at the top
+            # level of `parameters` (HassStartTimer / HassDecreaseTimer).
+            # Subtractive, so gated — Ollama honours the union and keeps it,
+            # and the Gemini pass below needs the anyOf intact to restate it.
             if provider in _FLATTEN_UNION_PROVIDERS:
                 flattened = _flatten_top_level_union(parameters)
                 if flattened is not parameters:
@@ -1922,7 +1929,7 @@ async def _retrieve_tools(  # noqa: PLR0915
             )
 
     # 4. Format and deduplicate. The provider drives the subtractive schema
-    # normalisation passes: the OpenAI top-level-union flatten
+    # normalisation passes: the OpenAI/Anthropic top-level-union flatten
     # (_flatten_top_level_union) and the Gemini anyOf-required sanitizer
     # (_sanitize_any_of_required).
     provider_raw = opts.get(CONF_CHAT_MODEL_PROVIDER)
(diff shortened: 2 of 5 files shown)


a4bell64:
Thanks for turning this around so fast, and for actually checking the schemas rather than taking my "all 25" at face value — that was wrong and I'd have gone on believing it.

I'd already run your finding in production before I saw the merge. On HA 2026.8.3 — one patch newer than the 2026.8.2 you converted against — I cut `tool_exclusions` from 25 down to just the seven timer tools, leaving the 18 action tools bound. No 400s: `HassTurnOn` / `HassTurnOff` execute cleanly, and history and live-context queries answer normally. Verified out loud from a Voice PE puck, not only over the text path. So the two-tool finding holds on 2026.8.3 as well.

One thing that might be worth a line in the test notes: in my install the seven timer tools show as "(not currently available)" in the options-flow dropdown and never enter the bound set on the text path — they only appear when the request comes from a timer-capable satellite. That's why this hid through a full day of typed testing and surfaced on my first spoken command. Anyone reproducing it purely via `conversation.process` won't see it.

Upgrading to v3.33.2 now and dropping the last seven exclusions. I'll report back if anything turns up.

PR #611, opened 2026-09-06
Add network security and privacy audit design plan

goruck:
Design proposal for a capability-based network security feature: a normalized network snapshot section populated by per-integration adapters (HA-native, generic router tracker, eero), capability-gated Sentinel rules, a known-device inventory, baseline and discovery extensions, a read-only audit_home_security agent tool, pseudonymization with an optional local model override, and a phased path to PIN-gated remediation.

Claude-Session: https://claude.ai/code/session_015yZmrSfWymqALXsWZLuVE6

### DIFF
--- a/docs/network-security-plan.md
+++ b/docs/network-security-plan.md
@@ -32,9 +32,9 @@ the installed integrations can see.
   alarms exposed to voice assistants, cloud remote access, pending updates).
 - Explain findings and per-device privacy exposure in plain language via the
   LLM, using pseudonymized identifiers and an optional local-model override.
-- Work on every install: a home with no router integration still gets the
-  HA-native audit; a home with eero, UniFi, or another supported router gets
-  progressively more.
+- Work on every install: a home with no router integration still gets a
+  complete HA hardening audit (see HA-Only MVP); a home with Fritz!Box,
+  UniFi, eero, or another supported router gets progressively more.
 - Reuse existing Sentinel machinery (rules, baseline, discovery, notifier,
   audit trail, triage) rather than build a parallel pipeline.
 
@@ -168,6 +168,14 @@ class HaSecurityPosture(TypedDict, total=False):
     pending_updates: list[str]  # update.* entity_ids
     http_use_x_forwarded_for: bool
     http_trusted_proxies_configured: bool
+    http_ip_ban_enabled: bool
+    trusted_networks_bypass_login: bool
+    addons_with_host_ports: dict[str, list[int]]  # slug -> host ports
+    addons_unprotected: list[str]  # slugs with protection mode off
+    webhook_automations_public: list[str]  # automation entity_ids
+    unavailable_security_devices: dict[str, int]  # entity_id -> minutes
+    discovered_unconfigured: list[dict[str, str]]  # {handler, source, title}
+    discovered_ignored: list[dict[str, str]]
 
 
 class NetworkSnapshot(TypedDict):
@@ -342,6 +350,25 @@ translation key or unique-id suffix, never by friendly name.
   security device when its device is a lock, alarm, camera, or router.
 - **Device-registry MAC index.** `connections` entries of type `mac`, used to
   join router clients to HA devices and their integration domain.
+- **Security device availability.** Every `lock`, `alarm_control_panel`, and
+  `camera` entity whose state is `unavailable`, with minutes since
+  `last_changed`. Needs no router; HA already knows when it lost a device.
+- **Discovered but unconfigured devices.** In-progress config flows whose
+  source is `ssdp`, `zeroconf`, `dhcp`, or `homekit`, plus config entries
+  with source `ignore`. This is a partial LAN inventory HA collects on
+  every install without any router integration.
+- **Supervisor add-ons** (HA OS and Supervised only). Via the `hassio`
+  component's add-on info: add-ons with container ports mapped to the host,
+  and add-ons running with protection mode off. Absent on Container and
+  Core installs, reported as a missing capability.
+- **Webhook automations.** Automations whose raw config has a webhook
+  trigger with `local_only: false`, read from the automation entities'
+  stored config. These are reachable from the internet whenever remote
+  access is on.
+- **Auth providers.** A `trusted_networks` provider with
+  `allow_bypass_login` enabled.
+- **HTTP settings.** `use_x_forwarded_for` without trusted proxies, and IP
+  banning disabled, from the `http` component's runtime configuration.
 
 ## Known-Device Inventory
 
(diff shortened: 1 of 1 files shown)


docs/network-security-plan.md
```diff
@@ -0,0 +1,861 @@
...
+any LLM call (explain, triage, discovery, or the tool result rendered for the
+agent) a `redact_network_identifiers()` step, next to `_redact_person_names()`
+in `explain/llm_explain.py`, strips `mac` and `ip`, replaces hostnames with
+`manufacturer + key`, and keeps only the subnet class of IPs ("lan",
+"public"). The notifier re-hydrates the display name from the inventory when
+rendering the final notification, after the model has produced its text.
+
+### Local-model override
+
+Add `network_audit` to `FEATURE_DEFS` (`{"name": "Network Audit", "required": False}`)
```
goruck, line 746: **[P1] Local override does not protect tool results**

The audit tool returns structured findings to the conversation graph, whose action node feeds the resulting `ToolMessage` back to the same conversation model. Adding a `network_audit` feature mapping therefore cannot keep audit data local while conversation uses a cloud provider. The design needs an explicit local-model invocation inside the tool and a precisely sanitized result for the conversation model, or it must disclose that the conversation provider receives the audit payload.

docs/network-security-plan.md
```diff
@@ -0,0 +1,861 @@
...
+  and the device-registry id for radio devices. Each row holds `source`
+  (`router`, `zigbee`, `zwave`, `bluetooth`, `matter`), `first_seen`,
+  `last_seen`, `trusted` (bool), the last observed name and manufacturer,
+  and the HA device id when known. Router MACs are one source among
+  several, not the inventory's identity.
+- **Bootstrap:** on the first run after enablement every currently connected
+  client is recorded with `trusted = True` and a single "inventory
+  established" notification asks the user to review it. This mirrors the
+  baseline establishment notification and avoids a flood of alerts on day one.
+- **Auto-trust:** a client that joins to an HA device in the registry is
```
goruck: **[P1] Do not auto-trust router-created registry devices**

A device-registry match does not establish user trust. For example, the Home Assistant FRITZ!Box integration creates `DeviceInfo` with a MAC connection for every tracked client, so every new client observed by that phase-1 adapter would immediately satisfy this auto-trust rule and bypass `network_unknown_device_joined`. Auto-trust must require evidence from a non-router integration or explicit user approval.

docs/network-security-plan.md
```diff
@@ -0,0 +1,861 @@
...
+| `counters.network.client_count` | count of connected trackers | baseline input |
+
+Not available from eero and therefore never asserted by the adapter: per-client
+guest flag, per-client blocked flag, port forwards, DHCP reservations, band or
+SSID, device type. The entity is matched by `platform` plus the integration's
+translation key or unique-id suffix, never by friendly name.
+
+#### `ha_native` adapter details
+
+- **Auth.** `hass.auth.async_get_users()` gives `is_admin`, `is_active`, and
```
goruck, line 427: **[P1] Persist the auth history used by change rules**

The current auth objects only provide present token and IP state, while the existing audit store retains a snapshot hash rather than the prior snapshot. No proposed store records users, token identifiers, or previously seen IPs, so `refresh_tokens_from_new_ip` and `ha_new_admin_or_token` cannot work reliably, especially after restart. Add a bootstrapped, persistent, non-secret auth inventory with explicit retention and deletion behavior.

docs/network-security-plan.md
```diff
@@ -0,0 +1,861 @@
...
+| `eero` | `platform == "eero"` | everything in the eero table below | 1 |
+| `zwave` | `platform == "zwave_js"` | per-node security class, inclusion state, controller update | 1 |
+| `zigbee` | `platform == "zha"` or Zigbee2MQTT bridge entities via `mqtt` | permit-join state, coordinator update, device list | 1 |
+| `radio_registry` | always | new radio devices from the device registry, per protocol | 1 |
+| `bluetooth` | the `bluetooth` integration is loaded | persistent unknown tracker-class advertisements | 2 |
+| `matter` | `platform == "matter"` | per-node fabric list | 3 |
+| `pi_hole` | `platform == "pi_hole"` | protection switch, blocked-query counters | 3 |
+| `asuswrt` / `tplink_omada` / `keenetic_ndms2` / `mikrotik` / `netgear` | by platform | clients; per-integration posture where exposed | 3 |
+
+Adapters must not import integration code. They read entity state and
```
goruck: **[P1] Entity-only adapters cannot supply the radio data**

This adapter boundary conflicts with the phase-1 radio design below. In the pinned Home Assistant runtime, Z-Wave `highest_security_class` and `inclusion_state` come from integration runtime node/controller objects exposed through the `zwave_js` WebSocket implementation, not entity states or attributes; the later plan itself says to read node objects and the ZHA gateway. Explicitly permit and version-test integration runtime APIs, or defer rules whose capabilities cannot be obtained through the declared boundary.

docs/network-security-plan.md
```diff
@@ -0,0 +1,861 @@
...
+| guest network idle | turn off | `switch.*_guest_network` |
+| WPA3 disabled | turn on | `switch.*_wpa3` (advisory, may drop old devices) |
+| protection disabled | turn on | `switch.*_advanced_security`, `switch.*_ad_blocking` |
+| router update pending | install | `update.install` |
+| stale long-lived token | revoke | HA auth API, via a new HGA service |
+| exposed sensitive entity | unexpose | exposed-entities helper, via a new HGA service |
+
+Mechanics:
+
+- `suggested_actions` becomes a list of `{label, domain, service, entity_id}`
```
goruck: **[P2] Preserve the suggested_actions string contract**

`AnomalyFinding.suggested_actions` is currently `list[str]`; notification prompts join its values, auto-execution calls `split` on each value, event payloads expose the string list, and the proposals UI filters for strings. Replacing it with descriptor dictionaries would break those consumers. Add a separate backward-compatible descriptor field, or include every producer, consumer, persisted record, event schema, and migration in the phase-2 plan.

goruck (commented):
Contract-level review of the network security and privacy design. The inline findings identify three security/privacy gaps and two implementation-contract gaps that should be resolved before this plan guides implementation.

docs/network-security-plan.md
goruck: Agreed. The Local-model override section now states that the `ToolMessage` always reaches the conversation provider and designs around it: the tool receives the `network_audit` model via `configurable` (same path as `vlm_model`), runs the narrative in-tool over the full pseudonymized payload, and returns only a digest (severity counts, narrative, capability lists, per-finding type/severity/pseudonymized key). `include_devices` is honored only when the conversation provider is the audit provider. When no `network_audit` feature is configured, the doc and README say plainly that the conversation provider sees the pseudonymized audit. A test asserts the digest key set.

---
_Generated by [Claude Code](https://claude.ai/code)_

docs/network-security-plan.md
goruck: Agreed. Auto-trust now requires that the joined registry device has at least one config entry whose domain is not a router, tracker, or discovery platform, and that the entry owns at least one non-`device_tracker` entity. A Shelly plug qualifies; a Fritz- or UniFi-created client device does not. Everything else stays untrusted until the user approves it. The inventory tests gain explicit negative cases for Fritz- and UniFi-created devices.

---
_Generated by [Claude Code](https://claude.ai/code)_

docs/network-security-plan.md
goruck: Agreed. Added an "Auth inventory store" section: `sentinel/auth_inventory.py` persists per-user id/admin/active/first_seen and per-token HA token id, type, client name, created_at, first_seen, and a bounded list of HMAC-pseudonymized `seen_ips`. No raw IPs, token values, or keys are stored. It bootstraps silently on first run, deletes token rows when HA drops the token, bounds `seen_ips` to 20 entries with `sentinel_auth_ip_retention_days` (default 90), is deleted with the Sentinel subentry, and has a reset service. Restart does not re-bootstrap. It is now step 3 in the implementation order, and tests cover detection across a simulated restart.

---
_Generated by [Claude Code](https://claude.ai/code)_

docs/network-security-plan.md
goruck: Agreed. Adapters are now split into two declared tiers. Entity adapters (all router and DNS adapters) keep the state-only boundary. Runtime adapters (`zwave`, ZHA half of `zigbee`, `bluetooth`, `matter`) may read integration runtime objects through the integration's own helper API under four rules: lazy import inside the function, every read degrades to a missing capability on ImportError/AttributeError/KeyError/TypeError, a test that exercises the real objects from the pinned `homeassistant` in `requirements/test.txt`, and a docstring naming the HA version and objects read. A runtime adapter that cannot get its data on the pinned version defers its rules rather than approximating from entity state.

---
_Generated by [Claude Code](https://claude.ai/code)_

docs/network-security-plan.md
goruck: Agreed. `suggested_actions` stays `list[str]`. Fixes go in a new additive `remediations: list[Remediation]` field with a default-empty factory, serialized under its own key in `as_dict()`, loaded as empty for pre-existing audit records via `_migrate_record`, and excluded from `build_anomaly_id()` like the display-only keys. The notifier renders `remediations` as buttons and keeps using `suggested_actions` for text. Each remediation is also mirrored into `suggested_actions` in the existing `domain.service` string form so the `allowed_services` allowlist and execution gate work unchanged. Existing rules are untouched.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #630, opened 2026-09-12
Fix LangChain tool schema extraction for injected arguments

MaestroMetty:
## Summary

Fixes model-facing schema generation for LangChain tools that use LangGraph-injected arguments (`store`, `config`).

Those fields are injected at runtime and must not be advertised to the model.

## Bug

HGA built tool JSON schemas with `args_schema.schema()`.

For tools such as `upsert_memory` and `confirm_sensitive_action`, that schema includes:

- `store: InjectedStore()`
- `config: InjectedToolArg()`

Pydantic cannot serialize `BaseStore`:

```text
Cannot generate a JsonSchema for core_schema.IsInstanceSchema
(<class 'langgraph.store.base.BaseStore'>)
```

On the fallback path this could bind an empty or invalid schema.

## Root cause

Local-tool indexing already preferred `tool_call_schema`, which excludes injected arguments.

`_get_fallback_tools()` in `graph.py` still used `args_schema.schema()`. That path runs when the tool index is unavailable, not ready, empty, or when `_get_tool_by_name` misses.

`conversation.py` had a separate extractor, so indexing and fallback could emit different schemas.

### `87a8a73`

Shared helper `langchain_tool_parameters_json()`:

- prefers `tool_call_schema`
- uses `args_schema` only when `tool_call_schema` is absent

Used by both `_get_fallback_tools()` and local-tool indexing.

Runtime is unchanged: LangGraph still injects `store` and `config`; they are no longer exposed to the model.

### `f296095`

Regression tests for `upsert_memory`, `confirm_sensitive_action`, ordinary tools, and fallback schema generation.

They check that injected fields are omitted, public parameters remain, and no `BaseStore` / `IsInstanceSchema` warning is raised.

## Expected result

Injected-arg tools expose only public parameters. Example for `upsert_memory`: `content`, `context`, `memory_id`. `store` and `config` stay runtime-only.


goruck:
Thanks for this — the bug is real and the diagnosis is exactly right. I reproduced it before reviewing:

```
upsert_memory            RAISED PydanticInvalidForJsonSchema: IsInstanceSchema (<class 'langgraph.store.base.BaseStore'>)
confirm_sensitive_action RAISED PydanticInvalidForJsonSchema: IsInstanceSchema (<class 'langgraph.store.base.BaseStore'>)
alarm_control            OK
...the other 7 local tools  OK
```

So `_get_fallback_tools()` set `parameters="{}"` for exactly those two, and `_format_and_dedupe_tools()` normalizes `{}` to `{"type":"object","properties":{}}` — meaning the model was told `confirm_sensitive_action` (the critical-action PIN gate) and `upsert_memory` take **no arguments**. That fires on the "vector store unavailable or index not yet ready" path and on `_get_tool_by_name()`'s index-miss path, i.e. at startup before indexing completes or when Postgres is down.

It's also been broken longer than it looks: the indexing path was fixed in `7607753` (2026-05-11), but the fallback added in `4122d75` (2026-04-09) was never brought along. Unifying both onto one helper is the right call.

### Verification

- Output is **byte-identical** to the old indexing path for all 10 local tools → no tool-index content-hash churn, no re-embedding storm on upgrade.
- Fallback schemas for the 8 healthy tools are identical to before; only the 2 broken ones change.
- End-to-end through `_format_and_dedupe_tools()` for anthropic / gemini / ollama: `['action_id','pin']` and `['content','context','memory_id']`.

### What I pushed to your branch

Heads up — I rebased and force-pushed to your fork's `main` (maintainer-edits was enabled). Your two commits are preserved as-is; my changes are a separate third commit. The rebase resolved one conflict: `graph.py`'s import block now also carries `is_on_off_intent` / `is_tool_excluded` from #632.

Five follow-ups, none of which change the shape of your fix:

1. **`make typecheck` regressed 0 → 1 error.** The test's config literal was `dict[str, Any]` where `_get_fallback_tools()` declares `RunnableConfig`. Now typed via a `TYPE_CHECKING` import. Worth knowing: CI runs Ruff, Pytest, Hassfest and HACS — not pyright — so this one only shows up in `make typecheck`.

2. **The guard didn't cover the property access.** `getattr(lc_tool, "tool_call_schema", None)` *calls* a computed property (`get_input_schema()` + `_create_subset_model()`), and `getattr`'s default only swallows `AttributeError`. Anything else it raised escaped into `_get_fallback_tools()`, which has no guard of its own — so one hostile schema would cost the turn every *other* fallback tool, where the old code cost one tool its parameters. The extraction now sits inside the existing `_SCHEMA_ERRORS` catch.

3. **`{}` is now logged.** It's indistinguishable downstream from a tool that genuinely takes no arguments, and `_format_and_dedupe_tools()` only warns when the stored JSON is *unparseable* — `{}` parses fine. That silence is precisely why this survived four months, so it now emits a warning naming the tool.

4. **Memoized per tool object.** `tool_call_schema` is an *uncached* property that rebuilds a Pydantic subset model on every read — measured 16.4 ms vs 5.7 ms for `args_schema.schema()` across the ten local tools (~2.9x), multiplied by up to `TOOL_INCLUSIONS_MAX_PER_TURN` on the miss path that exists *because* rebuilding local schemas per lookup is wasteful. `BaseTool` is a Pydantic model, so it's neither hashable nor a valid `WeakKeyDictionary` key; the memo keys on `id()` and stores a weakref proving the id still names the recorded object, with a finalizer so per-turn tool objects can't leak. Net: **16.45 ms → 0.05 ms**, below the pre-change cost.

5. **Test nits.** Collapsed `_schema_and_properties`/`_properties`, and dropped `"BaseStore"` from `_INJECTED_KEYS` — it can never be a property name; the real check is the raw-string assertion.

### One thing about the tests

`test_schema_extraction_does_not_warn_on_injected_store` and the warning assertion in `test_fallback_tools_use_tool_call_schema` are vacuous as regression guards — the old code *raised and was swallowed*, it never warned, so both pass on pre-PR code too. The assertions that actually pin the regression are `assert "content" in upsert_props` / `"action_id" in confirm_props`, and those do fail on `main`. I left them in place, just don't rely on them.

I added five tests and mutation-tested each one (reverting the corresponding fix makes it fail), so they're real guards.

**Status:** 3369 passed · Ruff clean · pyright 0 errors · CI 5/5 green. Ready to merge — thanks again for the careful write-up in the PR description, it made this quick to verify.

